PR #3, opened 2021-12-12
feat: added ObjectValidator

kyranet:
Added `s.object` and its utilities


src/index.ts
```diff
@@ -1,7 +1,10 @@
-import { Shapes } from './lib/Shapes';
+import { Shapes, Type } from './lib/Shapes';
 
 export const s = new Shapes();
+export type { Type };
```
favna, line 4: This has to be moved to `type-exports` after #4

src/lib/Shapes.ts
```diff
@@ -83,3 +88,5 @@ export class Shapes {
 		return new SetValidator(validator);
 	}
 }
+
+export type Type<V> = V extends BaseValidator<infer T> ? T : never;
```
favna, line 92: This should be moved to `util-types` after #4

src/tsconfig.json
```diff
@@ -3,7 +3,9 @@
 	"compilerOptions": {
 		"rootDir": "./",
 		"outDir": "../dist",
-		"composite": true
+		"composite": true,
+		"preserveConstEnums": true,
```
favna, line 7: If merging #4 into here doesn't auto fix it then this property should be removed as it's already set in @sapphire/ts-config

src/validators/BaseValidator.ts
```diff
@@ -62,6 +62,12 @@ export abstract class BaseValidator<T> {
 	}
 }
 
+export type Partialize<V> = V extends UnionValidator<infer T>
+	? UnionValidator<T | undefined>
+	: V extends BaseValidator<infer T>
+	? UnionValidator<T | undefined>
+	: never;
```
favna, line 69: After #4 this should be moved to `util-types`

src/validators/ObjectValidator.ts
```diff
@@ -0,0 +1,149 @@
...
+
+		return errors.length === 0 //
+			? Result.ok(finalResult)
+			: Result.err(new AggregateError(errors, 'Failed to match at least one of the properties'));
+	}
+}
+
+// eslint-disable-next-line @typescript-eslint/ban-types
+export type NonNullObject = {} & object;
+export type MappedObjectValidator<T> = { [key in keyof T]: BaseValidator<T[key]> };
```
favna, line 144: After #4 this should be moved to `util-types`

src/tsconfig.json
```diff
@@ -3,7 +3,9 @@
...
+		"useDefineForClassFields": false
```
favna, line 8: This line should be removed because it's set in the base config

favna (changes requested):

vladfrangu (dismissed):

src/lib/util-types.ts
```diff
@@ -1,5 +1,13 @@
...
  * @internal
  * This type is only used inside of the library and is not exported on the root level
  */
 export type Constructor<T> = (new (...args: readonly any[]) => T) | (abstract new (...args: readonly any[]) => T);
+
+export type Type<V> = V extends BaseValidator<infer T> ? T : never;
+
+// eslint-disable-next-line @typescript-eslint/ban-types
+export type NonNullObject = {} & object;
+export type MappedObjectValidator<T> = { [key in keyof T]: BaseValidator<T[key]> };
```
favna, line 13:
```suggestion
/**
 * @internal
 * This type is only used inside of the library and is not exported on the root level
 */
export type Type<V> = V extends BaseValidator<infer T> ? T : never;

/**
 * @internal
 * This type is only used inside of the library and is not exported on the root level
 */
// eslint-disable-next-line @typescript-eslint/ban-types
export type NonNullObject = {} & object;

/**
 * @internal
 * This type is only used inside of the library and is not exported on the root level
 */
export type MappedObjectValidator<T> = { [key in keyof T]: BaseValidator<T[key]> };
```

This will ensure they are skipped over by typedoc

favna (changes requested):

favna (approved):

src/lib/util-types.ts
vladfrangu: All of these types are literally exported from type export tho Favna..

favna (dismissed):

favna (approved):

vladfrangu (approved):

PR #17, opened 2022-01-23
feat: added number transformers

kyranet:
Adds the following methods:

- `BaseValidator#transform`
- `s.number.divisibleBy(n)`
- `s.number.abs`
- `s.number.sign`
- `s.number.trunc`
- `s.number.floor`
- `s.number.fround`
- `s.number.round`
- `s.number.ceil`

### DIFF
--- a/src/constraints/NumberConstraints.ts
+++ b/src/constraints/NumberConstraints.ts
@@ -98,3 +98,21 @@ export const numberNeNaN: IConstraint<number> = {
 			: Result.ok(input);
 	}
 };
+
+export function numberDivisibleBy(divider: number): IConstraint<number> {
+	const expected = `% ${divider}`;
+	return {
+		run(input: number) {
+			return input % divider === 0 //
+				? Result.ok(input)
+				: Result.err(
+						new ConstraintError(
+							'numberDivisibleBy',
+							`Expected number to be divisible by ${divider}, but received ${input}`,
+							input,
+							expected
+						)
+				  );
+		}
+	};
+}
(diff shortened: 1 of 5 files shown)


tests/validators/number.test.ts
```diff
@@ -1,5 +1,9 @@
 import { ConstraintError, s, ValidationError } from '../../src';
 
+const safeInteger = 42;
+// eslint-disable-next-line @typescript-eslint/no-loss-of-precision
+const unsafeInteger = 242043489611808769;
```
vladfrangu, line 5: Shouldn't this use MAX_SAFE_INTEGER or w/e

vladfrangu (approved):

tests/validators/number.test.ts
kyranet: No, because it's *un*safe

src/constraints/NumberConstraints.ts
```diff
@@ -98,3 +98,21 @@ export const numberNeNaN: IConstraint<number> = {
...
+			return input % divider === 0 //
+				? Result.ok(input)
+				: Result.err(
+						new ConstraintError(
+							'numberDivisibleBy',
+							`Expected number to be divisible by ${divider}, but received ${input}`,
+							input,
+							expected
+						)
+				  );
```
favna, line 115: I spy with my little eye and I see mixed tabs and spaces

image

favna (changes requested):

src/constraints/NumberConstraints.ts
kyranet: Do you seriously think I have the time to set 2 spaces after a bunch of tabs?

That is obviously prettier's doing

favna (approved):

PR #20, opened 2022-01-23
feat: add RecordValidator

Khasms:


vladfrangu (approved):

.gitattributes
```diff
@@ -0,0 +1 @@
+* text eol=lf
```
kyranet, line 1: Please revert.

src/validators/RecordValidator.ts
```diff
@@ -0,0 +1,36 @@
...
+		super(constraints);
+		this.validator = validator;
+	}
+
+	protected override clone(): this {
+		return Reflect.construct(this.constructor, [this.validator, this.constraints]);
+	}
+
+	protected handle(value: unknown): Result<Record<string, T>, ValidationError | AggregateError> {
+		if (!(typeof value === 'object')) {
```
kyranet, line 19:
```suggestion
		if (typeof value !== 'object') {
```

src/validators/RecordValidator.ts
```diff
@@ -0,0 +1,36 @@
...
+			return Result.err(new ValidationError('RecordValidator', 'Expected a record', value));
+		}
+
+		const errors: Error[] = [];
```
kyranet, line 23: Add check for null, see ObjectValidator.

src/validators/RecordValidator.ts
```diff
@@ -0,0 +1,36 @@
...
+		this.validator = validator;
+	}
+
+	protected override clone(): this {
+		return Reflect.construct(this.constructor, [this.validator, this.constraints]);
+	}
+
+	protected handle(value: unknown): Result<Record<string, T>, ValidationError | AggregateError> {
+		if (!(typeof value === 'object')) {
+			return Result.err(new ValidationError('RecordValidator', 'Expected a record', value));
```
kyranet, line 20: Shouldn't this say it expects an object? JS-only users will be more familiar with "object" than with this TypeScript construct.

src/validators/RecordValidator.ts
kyranet: Also add a test for it.

kyranet (changes requested):

src/validators/RecordValidator.ts
Khasms: I figured if they didn't know what a record was, they wouldn't be using the record validator, so most people who came across this error would know what a record was and it would help distinguish it from the object validator a bit more.

.gitattributes
Khasms: Is there a reason for this? Having to convert the EOLs on every file all the time is a hassle.

kyranet: It should be a different PR.

src/validators/RecordValidator.ts
kyranet: Problem is, libraries may use this library, and they may use record, something developers using said library wouldn't see.

kyranet (approved):

vladfrangu (approved):

favna (approved):

favna:
@all-contributors please add @Khasms  for code

PR #21, opened 2022-01-24
feat: add MapValidator

Khasms:


src/validators/MapValidator.ts
```diff
@@ -0,0 +1,40 @@
...
+
+		const errors: Error[] = [];
+		const transformed = new Map<K, V>();
+
+		for (const [key, val] of value.entries()) {
+			const keyResult = this.keyValidator.run(key);
+			const valueResult = this.valueValidator.run(val);
+			const results = [keyResult, valueResult].filter((result) => result.isErr());
+			if (results.length === 0) transformed.set(keyResult.value!, valueResult.value!);
+			else errors.push(...results.map((result) => result.error!));
```
kyranet, line 33: This could be rewritten to not use two array allocations, even if it's a bit messy, we're trying to maximise performance in this library 😅

tests/validators/map.test.ts
```diff
@@ -0,0 +1,35 @@
...
+	test('GIVEN a matching map THEN returns a map', () => {
+		expect(predicate.parse(value)).toStrictEqual(value);
+	});
+
+	test('GIVEN a non-matching map THEN throws AggregateError', () => {
+		const map = new Map();
+		map.set('a', 1);
+		map.set('foo', 'bar');
+		map.set(2, 'fizz');
+		map.set(3, 'buzz');
```
kyranet, line 22: You can also do
```ts
const map = new Map([
	['a', 1],
	['foo', 'bar'],
	[2, 'fizz'],
	[3, 'buzz']
]);
```

Also you should add a line for:

- 2 valid
- key invalid
- value invalid
- 2 invalid

That way we can check that it's adding both keys

src/validators/MapValidator.ts
kyranet: I was thinking something among the lines of this:
```typescript
const length = errors.length;
if (keyResult.isErr()) errors.push(keyResult.error);
if (valueResult.isErr()) errors.push(valueResult.error);
if (errors.length === length) transformed.set(keyResult.value!, valueResult.value!);
```

Which works because if one of them error, the array's length is increased, and therefore doesn't match the length before pushing the key.

kyranet (changes requested):

tests/validators/map.test.ts
Khasms: Setting it in the constructor throws a TS error because of the mismatched types, so I can either `@ts-ignore` it, or stick with `set`.

favna (dismissed):

kyranet (approved):

favna (approved):

PR #23, opened 2022-01-26
build: add `.gitattributes`

Khasms:


.gitattributes
```diff
@@ -0,0 +1 @@
+* text eol=lf
```
kyranet, line 1:
```suggestion
* text eol=lf

```

kyranet (changes requested):

kyranet (approved):

favna:
As per comment above I am closing this

PR #25, opened 2022-01-26
feat: add `default`

Khasms:


kyranet:
This is not the right approach, the right approach is to make a DefaultValidator taking a BaseValidator. When handle receives `undefined`, it'll get the default, otherwise it'll run the underlying validator.

This way validators will suffer no overhead from defaults.

favna (dismissed):

Khasms:
> This is not the right approach, the right approach is to make a DefaultValidator taking a BaseValidator. When handle receives `undefined`, it'll get the default, otherwise it'll run the underlying validator. `.default` would then do `new DefaultValidator(this);`, and this isn't exported in Shapes.
> 
> This way validators will suffer no overhead from defaults.

I knew I missed something.

src/validators/DefaultValidator.ts
```diff
@@ -0,0 +1,26 @@
+import type { IConstraint } from '../constraints/base/IConstraint';
+import type { ValidationError } from '../lib/errors/ValidationError';
+import { Result } from '../lib/Result';
+import { BaseValidator } from './imports';
+import { getValue } from './util/getValue';
+
+export class DefaultValidator<T> extends BaseValidator<T> {
```
kyranet, line 7:
```suggestion
export class DefaultValidator<T> extends BaseValidator<T | undefined> {
```

src/validators/DefaultValidator.ts
```diff
@@ -0,0 +1,26 @@
...
+	private readonly validator: BaseValidator<T>;
+	private readonly defaultValue: T | (() => T);
+
+	public constructor(validator: BaseValidator<T>, value: T | (() => T), constraints: readonly IConstraint<T>[] = []) {
+		super(constraints);
+		this.validator = validator;
+		this.defaultValue = value;
+	}
+
+	protected handle(value: undefined): Result<T, ValidationError | AggregateError> {
```
kyranet, line 17:
```suggestion
	protected handle(value: T | undefined): Result<T, ValidationError | AggregateError> {
```

kyranet (changes requested):

src/validators/DefaultValidator.ts
Khasms: This was actually a mistype, I meant to put unknown instead of undefined, should I change it to unknown to match every other validator or use this suggestion?

kyranet: unknown, I updated the suggestion.

kyranet (approved):

favna (approved):

PR #32, opened 2022-02-09
feat: add bigint methods

Khasms:


src/validators/BigIntValidator.ts
```diff
@@ -37,6 +37,22 @@ export class BigIntValidator<T extends bigint> extends BaseValidator<T> {
 		return this.lt(0n);
 	}
 
+	public divisibleBy(number: bigint): this {
+		return this.addConstraint(bigintDivisibleBy(number) as IConstraint<T>);
+	}
+
+	public get abs(): this {
+		return this.transform(((value) => (value < 0 ? -value : value)) as (value: bigint) => T);
```
kyranet, line 45:
```suggestion
		return this.transform((value) => (value < 0 ? -value : value) as T);
```

src/validators/BigIntValidator.ts
```diff
@@ -37,6 +37,22 @@ export class BigIntValidator<T extends bigint> extends BaseValidator<T> {
...
+	}
+
+	public intN(bits: number): this {
+		return this.transform(((value) => BigInt.asIntN(bits, value)) as (value: bigint) => T);
```
kyranet, line 49:
```suggestion
		return this.transform((value) => BigInt.asIntN(bits, value) as T);
```

src/validators/BigIntValidator.ts
```diff
@@ -37,6 +37,22 @@ export class BigIntValidator<T extends bigint> extends BaseValidator<T> {
...
+	}
+
+	public uintN(bits: number): this {
+		return this.transform(((value) => BigInt.asUintN(bits, value)) as (value: bigint) => T);
```
kyranet, line 53:
```suggestion
		return this.transform((value) => BigInt.asUintN(bits, value) as T);
```

tests/validators/bigint.test.ts
```diff
@@ -0,0 +1,178 @@
+import { ConstraintError, s, ValidationError } from '../../src';
+
+const safeInteger = 42n;
+// eslint-disable-next-line @typescript-eslint/no-loss-of-precision
+const unsafeInteger = 242043489611808769n;
```
kyranet, line 5: They're always safe, and the eslint ignore is unneeded.

kyranet (changes requested):

favna (approved):

PR #35, opened 2022-02-13
feat: display the property that errored

kyranet:
Now we have a proper way to identify which properties failed to be validated

image


favna (dismissed):

src/validators/ArrayValidator.ts
```diff
@@ -17,22 +17,22 @@ export class ArrayValidator<T> extends BaseValidator<T[]> {
...
 			return Result.err(new ValidationError('ArrayValidator', 'Expected an array', values));
 		}
 
-		const errors: BaseError[] = [];
+		const errors: [number, BaseError][] = [];
 		const transformed: T[] = [];
 
-		for (const value of values) {
-			const result = this.validator.run(value);
+		for (const entry of values.entries()) {
```
vladfrangu, line 28: Shouldn't we use normal for-i loops for speed?

tests/lib/errors/CombinedPropertyError.test.ts
```diff
@@ -0,0 +1,61 @@
...
+				'  Received one or more errors',
+				'',
+				'  .foo',
+				'  | ValidationError > StringValidator',
+				'  |   Expected a string primitive',
+				'  | ',
+				'  |   Received:',
+				'  |   | 42',
+				'',
+				'  [2]',
```
vladfrangu, line 33: [2] feels weird.. is it input[2]? Is it just the value 2?

vladfrangu (changes requested):

tests/lib/errors/CombinedPropertyError.test.ts
kyranet: `input[2]`

vladfrangu (approved):

favna (approved):

PR #38, opened 2022-02-24
feat: add remaining string validations

imranbarbhuiya:
This pr adds `url`, `uuid`, `regex` methods in string validator.


kyranet:
You did your benchmark wrong, the RegExp instantiation should happen in the setup, not in every benchmark cycle.

Moving it to the benchmark cycle makes the RegExp almost 3 times faster, however, I'm not accepting the RegExp.

src/validators/StringValidator.ts
```diff
@@ -29,6 +37,28 @@ export class StringValidator<T extends string> extends BaseValidator<T> {
...
 	}
 
+	public email(): this {
+		const emailRegex = /^(([^<>()[\]\.,;:\s@\"]+(\.[^<>()[\]\.,;:\s@\"]+)*)|(\".+\"))@(([^<>()[\]\.,;:\s@\"]+\.)+[^<>()[\]\.,;:\s@\"]{2,})$/i;
+		return this.addConstraint(stringRegex(emailRegex, 'email') as IConstraint<T>);
+	}
+
+	public url(allowedProtocols: string[] = ['https?']): this {
+		const urlRegex = new RegExp(`^(?:${allowedProtocols.join('|')}):\/\/[^\s\.]+\.\\S{2,}$`);
+		return this.addConstraint(stringRegex(urlRegex, 'url') as IConstraint<T>);
```
kyranet, line 47: This is vulnerable to accidental ReDoS injection due to unsanitised user-defined RegExp, and security tools won't pick up this kind of stuff.

imranbarbhuiya:
> You did your benchmark wrong, the RegExp instantiation should happen in the setup, not in every benchmark cycle.
> 
> Moving it to the benchmark cycle makes the RegExp almost 3 times faster, however, I'm not accepting the RegExp.

So should I use a different regex or use `URL`?

src/validators/StringValidator.ts
kyranet: Also, it should validate using `new URL(input)`, not a RegExp, no matter how "slow" it is, it's the fastest option among full WHATWG-compliant URL parsers, and it's also built-in.

src/validators/StringValidator.ts
```diff
@@ -29,6 +37,28 @@ export class StringValidator<T extends string> extends BaseValidator<T> {
 		return this.addConstraint(stringLengthNe(length) as IConstraint<T>);
 	}
 
+	public email(): this {
+		const emailRegex = /^(([^<>()[\]\.,;:\s@\"]+(\.[^<>()[\]\.,;:\s@\"]+)*)|(\".+\"))@(([^<>()[\]\.,;:\s@\"]+\.)+[^<>()[\]\.,;:\s@\"]{2,})$/i;
```
kyranet, line 41: Do you even know what this RegExp does? If so, are you aware that the URL validation in here doesn't match the RegExp you used in the url method?

It should at least have a link to the source where you got the RegExp from, and it should also be moved outside the method, there isn't a need to create a new RegExp instance over and over.

src/validators/StringValidator.ts
```diff
@@ -29,6 +37,28 @@ export class StringValidator<T extends string> extends BaseValidator<T> {
...
+	public url(allowedProtocols: string[] = ['https?']): this {
+		const urlRegex = new RegExp(`^(?:${allowedProtocols.join('|')}):\/\/[^\s\.]+\.\\S{2,}$`);
+		return this.addConstraint(stringRegex(urlRegex, 'url') as IConstraint<T>);
+	}
+
+	public uuid(v?: 1 | 2 | 3 | 4 | 5): this {
+		const uuidRegex = new RegExp(
+			`^([0-9A-F]{8}-[0-9A-F]{4}-[${v ?? '1-5'}][0-9A-F]{3}-[89AB][0-9A-F]{3}-[0-9A-F]{12}|00000000-0000-0000-0000-000000000000)$`,
+			'i'
+		);
```
kyranet, line 54: Are all UUID versions really the same format?

kyranet: Also, it should default to UUIDv4, since that's the most popular version.

wdyt? @favna @vladfrangu

src/validators/StringValidator.ts
```diff
@@ -29,6 +37,28 @@ export class StringValidator<T extends string> extends BaseValidator<T> {
...
+		const emailRegex = /^(([^<>()[\]\.,;:\s@\"]+(\.[^<>()[\]\.,;:\s@\"]+)*)|(\".+\"))@(([^<>()[\]\.,;:\s@\"]+\.)+[^<>()[\]\.,;:\s@\"]{2,})$/i;
+		return this.addConstraint(stringRegex(emailRegex, 'email') as IConstraint<T>);
+	}
+
+	public url(allowedProtocols: string[] = ['https?']): this {
+		const urlRegex = new RegExp(`^(?:${allowedProtocols.join('|')}):\/\/[^\s\.]+\.\\S{2,}$`);
+		return this.addConstraint(stringRegex(urlRegex, 'url') as IConstraint<T>);
+	}
+
+	public uuid(v?: 1 | 2 | 3 | 4 | 5): this {
```
kyranet, line 50: Doesn't support `${bigint}-${bigint}`, if we're keeping the code below.

tests/validators/string.test.ts
```diff
@@ -95,5 +95,86 @@ describe('StringValidator', () => {
...
+			test.each(['6e8bc430-9a1b-4f7f-b7a5-ea4dede09a4b'])('GIVEN %s THEN returns given value', (input) => {
+				expect(uuid4Predicate.parse(input)).toBe(input);
+			});
+
+			test.each(['6e8bc430-9a1b-4f7f-b7a5', '6e8bc430-9a1b-4f7f-b7a5'])('GIVEN %s THEN throws a ConstraintError', (input) => {
+				expect(() => uuid4Predicate.parse(input)).toThrow(
+					new ConstraintError('s.string.uuid', 'Invalid string format', input, 'expected.uuid')
+				);
+			});
+		});
```
kyranet, line 163: This should be a nested describe inside uuid, and named `v4`.

kyranet (changes requested):
Since we're adding `url`, we should probably add [`ip`](https://nodejs.org/dist/latest-v17.x/docs/api/net.html#netisipinput), [`ipv4`](https://nodejs.org/dist/latest-v17.x/docs/api/net.html#netisipv4input), and [`ipv6`](https://nodejs.org/dist/latest-v17.x/docs/api/net.html#netisipv6input), using the methods masked-linked in each name, and also as getters (we'll keep the new stuff from this PR as methods, since we can add allowed domains, protocols, etc, on top of the URL parser).

src/validators/StringValidator.ts
imranbarbhuiya: > Do you even know what this RegExp does? If so, are you aware that the URL validation in here doesn't match the RegExp you used in the url method?

I'd added it at last that's why I didn't changed the url regex. Ik this checks for all special characters in domain name but in url I'd only added check for space. I'll remove the url regex so no need to change that now.

src/validators/StringValidator.ts
imranbarbhuiya: > Are all UUID versions really the same format?

https://stackoverflow.com/a/38191078/16893382

I've tested with few uuids and it was working fine.

kyranet: Honestly, this should use `null` if you want `1-5`, otherwise default to `4`.

src/validators/StringValidator.ts
imranbarbhuiya: changed it also added allowedDomains.

src/validators/StringValidator.ts
imranbarbhuiya: changed

src/validators/StringValidator.ts
imranbarbhuiya: added

src/constraints/util/operators.ts
```diff
@@ -38,3 +38,8 @@ export interface Comparator {
 	(a: number, b: number): boolean;
 	(a: bigint, b: bigint): boolean;
 }
+
+export interface UrlOptions {
+	allowedProtocols?: string[];
```
kyranet, line 43: Should require `:` at the end of all strings, make this a `StringProtocol` type and export it:
```suggestion
	allowedProtocols?: `${string}:`[];
```

src/constraints/util/operators.ts
```diff
@@ -38,3 +38,8 @@ export interface Comparator {
...
+	allowedDomains?: string[];
```
kyranet, line 44: Should require `.` to be contained so invalid domains can't be passed, make this a `StringDomain` type and export it:
```suggestion
	allowedDomains?: `${string}.${string}`[];
```

src/constraints/StringConstraints.ts
```diff
@@ -44,3 +47,48 @@ export function stringLengthNe(length: number): IConstraint<string> {
...
+				if (options?.allowedDomains && !options.allowedDomains.includes(url.hostname)) {
+					return Result.err(
+						new ConstraintError(
+							's.string.url',
+							'Invalid URL domain',
+							input,
+							`expected.url.domain in ${options.allowedDomains.join(', ')}`
+						)
+					);
+				}
```
kyranet, line 86: I'm not a fan of doing those checks over and over, it slows down an already slow operation, not to mention it's also inside a try/catch, which makes performance worse.

kyranet (changes requested):

src/constraints/StringConstraints.ts
imranbarbhuiya: yeah sure. So should I remove them or keep them as it is?

src/validators/StringValidator.ts
```diff
@@ -29,6 +40,27 @@ export class StringValidator<T extends string> extends BaseValidator<T> {
...
+	}
+
+	public url(options?: UrlOptions): this {
+		return this.addConstraint(stringUrl(options) as IConstraint<T>);
+	}
+
+	public uuid(v: number | `${bigint}-${bigint}` | null = 4): this {
+		// from https://stackoverflow.com/a/38191078/16893382
+		const uuidRegex = new RegExp(
+			`^([0-9A-F]{8}-[0-9A-F]{4}-[${v ?? '1-5'}][0-9A-F]{3}-[89AB][0-9A-F]{3}-[0-9A-F]{12}|00000000-0000-0000-0000-000000000000)$`,
```
kyranet, line 54: Wait, why is there a null uuid?

src/constraints/StringConstraints.ts
```diff
@@ -3,7 +3,17 @@ import { Result } from '../lib/Result';
 import type { IConstraint } from './base/IConstraint';
 import { Comparator, eq, ge, gt, le, lt, ne } from './util/operators';
 
-export type StringConstraintName = `s.string.length${'Lt' | 'Le' | 'Gt' | 'Ge' | 'Eq' | 'Ne'}`;
+export type StringConstraintName = `s.string.${`length${'Lt' | 'Le' | 'Gt' | 'Ge' | 'Eq' | 'Ne'}` | 'regex' | 'url' | 'uuid' | 'email'}`;
+export type StringProtocol = `${string}:`;
+export type StringDomain = `${string}.${string}`;
+
+export interface UrlOptions {
```
kyranet, line 10: Those aren't exported, check for where `StringContraintName` is exported at.

kyranet (changes requested):

src/validators/StringValidator.ts
imranbarbhuiya: https://github.com/sapphiredev/shapeshift/pull/38#discussion_r813677328
ig I've misunderstood it.

kyranet: Yes, that comment was resolved correctly, `v` being `null` won't default to `4`, so it'll be defaulted to `1-5` in line 54.

However, a null UUID isn't... exactly what I meant, they're also not valid as well

src/constraints/StringConstraints.ts
imranbarbhuiya: wdym? I didn't understand.

src/validators/StringValidator.ts
```diff
@@ -29,6 +41,39 @@ export class StringValidator<T extends string> extends BaseValidator<T> {
...
+		);
+		return this.addConstraint(stringRegex(uuidRegex, 'uuid') as IConstraint<T>);
+	}
+
+	public regex(regex: RegExp): this {
+		return this.addConstraint(stringRegex(regex, 'regex') as IConstraint<T>);
+	}
+
+	public get ipv4(): this {
+		return this.addConstraint(stringIp(4) as IConstraint<T>);
```
kyranet, line 66:
```suggestion
		return this.ip(4);
```

src/validators/StringValidator.ts
```diff
@@ -29,6 +41,39 @@ export class StringValidator<T extends string> extends BaseValidator<T> {
...
+	}
+
+	public get ipv6(): this {
+		return this.addConstraint(stringIp(6) as IConstraint<T>);
```
kyranet, line 70:
```suggestion
		return this.ip(6);
```

kyranet (changes requested):

src/validators/StringValidator.ts
imranbarbhuiya: Oh, that `00` I've added the regex from the StackOverflow link but then as you'd asked that all 5 types of uuid have the same regex or not so I checked zod regex and found they have a null uuid. Idk why it's there and I though it's also a valid uuid so I added it. I'll remove it if u want.

src/constraints/StringConstraints.ts
kyranet: We can't do `import type { UrlOptions, StringProtocol, StringDomain } from '@sapphire/shapeshift';`, which is something we'd like to be able to since they're public-facing.

imranbarbhuiya: Thank u for the detailed explanation. Fixed

src/validators/StringValidator.ts
imranbarbhuiya: Now after searching a bit I found its a valid uuid.

Source https://en.m.wikipedia.org/wiki/Universally_unique_identifier
Screenshot_20220224-194256_Chrome

kyranet: Oh, interesting, I didn't know that, then it's fine.

src/constraints/StringConstraints.ts
```diff
@@ -2,8 +2,21 @@ import { ConstraintError } from '../lib/errors/ConstraintError';
 import { Result } from '../lib/Result';
 import type { IConstraint } from './base/IConstraint';
 import { Comparator, eq, ge, gt, le, lt, ne } from './util/operators';
+import net from 'net';
 
-export type StringConstraintName = `s.string.length${'Lt' | 'Le' | 'Gt' | 'Ge' | 'Eq' | 'Ne'}`;
+export type StringConstraintName =
+	| `s.string.${`length${'Lt' | 'Le' | 'Gt' | 'Ge' | 'Eq' | 'Ne'}` | 'regex' | 'url' | 'uuid' | 'email'}`
+	| StringIpName;
+type StringIpName = `s.string.ip${'v4' | 'v6' | ''}`;
```
vladfrangu, line 10: Inline these

src/constraints/StringConstraints.ts
```diff
@@ -2,8 +2,21 @@ import { ConstraintError } from '../lib/errors/ConstraintError';
...
+export type StringProtocol = `${string}:`;
+export type StringDomain = `${string}.${string}`;
+export interface UrlOptions {
+	allowedProtocols?: StringProtocol[];
+	allowedDomains?: StringDomain[];
+}
```
vladfrangu, line 16:
```suggestion
export type StringProtocol = `${string}:`;

export type StringDomain = `${string}.${string}`;

export interface UrlOptions {
	allowedProtocols?: StringProtocol[];
	allowedDomains?: StringDomain[];
}
```

src/constraints/StringConstraints.ts
```diff
@@ -2,8 +2,21 @@ import { ConstraintError } from '../lib/errors/ConstraintError';
...
+
+// from https://stackoverflow.com/a/46181/1550155
+export const emailRegex = /^(([^<>()[\]\.,;:\s@\"]+(\.[^<>()[\]\.,;:\s@\"]+)*)|(\".+\"))@(([^<>()[\]\.,;:\s@\"]+\.)+[^<>()[\]\.,;:\s@\"]{2,})$/i;
```
vladfrangu, line 19: Could we maybe use a module for this instead? thoughts @kyranet @favna ?

src/constraints/StringConstraints.ts
```diff
@@ -44,3 +57,68 @@ export function stringLengthNe(length: number): IConstraint<string> {
 	const expected = `expected.length !== ${length}`;
 	return stringLengthComparator(ne, 's.string.lengthNe', expected, length);
 }
+
+export function stringRegex(regex: RegExp, type: 'url' | 'uuid' | 'regex' | 'email'): IConstraint<string> {
+	const expected = `expected.${type}`;
```
vladfrangu, line 62:
```suggestion
	const expected = `expected to be a ${type}`;
```

src/constraints/StringConstraints.ts
```diff
@@ -44,3 +57,68 @@ export function stringLengthNe(length: number): IConstraint<string> {
...
+							'Invalid URL domain',
+							input,
+							`expected.url.domain in ${options.allowedDomains.join(', ')}`
+						)
+					);
+				}
+
+				return Result.ok(input);
+			} catch {
+				return Result.err(new ConstraintError('s.string.url', 'Invalid URL', input, 'expected.url'));
```
vladfrangu, line 102:
```suggestion
				return Result.err(new ConstraintError('s.string.url', 'Invalid URL', input, 'expected to be a valid url'));
```

src/constraints/StringConstraints.ts
```diff
@@ -44,3 +57,68 @@ export function stringLengthNe(length: number): IConstraint<string> {
...
+	return {
+		run(input: string) {
+			return (version === 4 ? net.isIPv4(input) : version === 6 ? net.isIPv6(input) : net.isIP(input)) //
+				? Result.ok(input)
+				: Result.err(
+						new ConstraintError(
+							`s.string.ip${ipVersion}` as StringIpName,
+							`Invalid ip${ipVersion} address`,
+							input,
+							`expected.ip${ipVersion}`
```
vladfrangu, line 119:
```suggestion
							`expected to be an ip${ipVersion} address`
```

tests/validators/string.test.ts
```diff
@@ -95,5 +95,138 @@ describe('StringValidator', () => {
 				);
 			});
 		});
+
+		describe('email', () => {
+			const emailPredicate = s.string.email;
+
+			test.each(['hi@hello.com', 'foo@bar.net'])('GIVEN %s THEN returns given value', (input) => {
```
vladfrangu, line 102:
```suggestion
			test.each(['hi@hello.com', 'foo@bar.net', 'hello+world@example.com'])('GIVEN %s THEN returns given value', (input) => {
```

vladfrangu (changes requested):

src/constraints/StringConstraints.ts
imranbarbhuiya: I'm using it again in line 118

src/constraints/StringConstraints.ts
kyranet: I just realised, this has a little small flaw.

- `expected to be a url` (should be `to match an URL`)
- `expected to be a uuid` (should be `to match an UUID`)
- `expected to be a regex` (???????, should be `to match ${pattern}`)
- `expected to be a email` (should be `to match an email`)

src/constraints/StringConstraints.ts
kyranet: If the module is a fully compliant parser, sure.
If the module is just RegExp#test, then we use the RegExp.

imranbarbhuiya: We can use 

https://www.npmjs.com/package/email-validator

src/constraints/StringConstraints.ts
imranbarbhuiya: updated

src/constraints/StringConstraints.ts
kyranet: Uhm... [the code](https://github.com/manishsaraan/email-validator/blob/master/index.js) could use a refactor for improved performance, please leave a TODO for me, I'll rewrite it to a faster implementation.

src/constraints/StringConstraints.ts
```diff
@@ -2,8 +2,23 @@ import { ConstraintError } from '../lib/errors/ConstraintError';
...
 import type { IConstraint } from './base/IConstraint';
 import { Comparator, eq, ge, gt, le, lt, ne } from './util/operators';
+import net from 'net';
 
-export type StringConstraintName = `s.string.length${'Lt' | 'Le' | 'Gt' | 'Ge' | 'Eq' | 'Ne'}`;
+export type StringConstraintName =
+	| `s.string.${`length${'Lt' | 'Le' | 'Gt' | 'Ge' | 'Eq' | 'Ne'}` | 'regex' | 'url' | 'uuid' | 'email'}`
+	| StringIpName;
+type StringIpName = `s.string.ip${'v4' | 'v6' | ''}`;
+export type StringProtocol = `${string}:`;
```
vladfrangu, line 11:
```suggestion
	| StringIpName;

export type StringIpName = `s.string.ip${'v4' | 'v6' | ''}`;

export type StringProtocol = `${string}:`;
```

src/constraints/StringConstraints.ts
```diff
@@ -44,3 +59,68 @@ export function stringLengthNe(length: number): IConstraint<string> {
 	const expected = `expected.length !== ${length}`;
 	return stringLengthComparator(ne, 's.string.lengthNe', expected, length);
 }
+
+export function stringRegex(regex: RegExp, type: 'url' | 'uuid' | 'regex' | 'email'): IConstraint<string> {
+	const expected = `expected to match ${type === 'regex' ? regex.source : `an ${type}`}`;
```
vladfrangu, line 64: Bleh, the only one that needs the `n` in `an` is email, everything else looks out of place... @kyranet @favna should we just bite the bullet on `a email` or...

src/constraints/StringConstraints.ts
```diff
@@ -44,3 +59,68 @@ export function stringLengthNe(length: number): IConstraint<string> {
...
+				const url = new URL(input);
+
+				// TODO: optimize theses checks
+				if (options?.allowedProtocols && !options.allowedProtocols.includes(url.protocol as StringProtocol)) {
+					return Result.err(
+						new ConstraintError(
+							's.string.url',
+							'Invalid URL protocol',
+							input,
+							`expected.url.protocol in ${options.allowedProtocols.join(', ')}`
```
vladfrangu, line 87:
```suggestion
							`expected ${url.protocol} to be one of: ${options.allowedProtocols.join(', ')}`
```

src/constraints/StringConstraints.ts
```diff
@@ -44,3 +59,68 @@ export function stringLengthNe(length: number): IConstraint<string> {
...
+						)
+					);
+				}
+				if (options?.allowedDomains && !options.allowedDomains.includes(url.hostname as StringDomain)) {
+					return Result.err(
+						new ConstraintError(
+							's.string.url',
+							'Invalid URL domain',
+							input,
+							`expected.url.domain in ${options.allowedDomains.join(', ')}`
```
vladfrangu, line 97:
```suggestion
							`expected ${url.hostname} to be one of: ${options.allowedDomains.join(', ')}`
```

tests/validators/string.test.ts
```diff
@@ -95,5 +95,154 @@ describe('StringValidator', () => {
...
+					);
+				});
+			});
+		});
+
+		describe('uuid', () => {
+			describe('default', () => {
+				const uuidPredicate = s.string.uuid();
+
+				test.each(['450d6a23-9e6f-45d9-9d5a-fd4f6e014f16'])('GIVEN %s THEN returns given value', (input) => {
```
vladfrangu, line 159: Should also test null uuid

tests/validators/string.test.ts
```diff
@@ -95,5 +95,154 @@ describe('StringValidator', () => {
...
+				test.each(['6e8bc430-9a1b-4f7f-b7a5-ea4dede09a4b', '6e8bc430-9a1b-1f7f-b7a5-ea4dede09a4b', '6e8bc430-9a1b-3f7f-b7a5-ea4dede09a4b'])(
+					'GIVEN %s THEN returns given value',
+					(input) => {
+						expect(uuidRangePredicate.parse(input)).toBe(input);
+					}
+				);
+
+				test.each(['6e8bc430-9a1b-5f7f-b7a5-ea4dede09a4b'])('GIVEN %s THEN throws a ConstraintError', (input) => {
+					expect(() => uuidRangePredicate.parse(input)).toThrow(
+						new ConstraintError('s.string.uuid', 'Invalid string format', input, 'expected.uuid')
```
vladfrangu, line 182: This error message is not clear at all...  Ideally it should say "expected UUID version to be in range ${range}" or similar

tests/validators/string.test.ts
```diff
@@ -95,5 +95,154 @@ describe('StringValidator', () => {
...
+			const regex = /^[a-z]+$/;
+			const regexPredicate = s.string.regex(regex);
+
+			test.each(['abc', 'xyz'])('GIVEN %s THEN returns given value', (input) => {
+				expect(regexPredicate.parse(input)).toBe(input);
+			});
+
+			test.each(['ABC', '123A'])('GIVEN %s THEN throws a ConstraintError', (input) => {
+				expect(() => regexPredicate.parse(input)).toThrow(
+					new ConstraintError('s.string.regex', 'Invalid string format', input, `expected.regex`)
```
vladfrangu, line 198: This message needs updating

tests/validators/string.test.ts
```diff
@@ -95,5 +95,154 @@ describe('StringValidator', () => {
...
+		describe('ip', () => {
+			describe('default', () => {
+				const ipPredicate = s.string.ip();
+
+				test.each(['::1', '127.0.0.1'])('GIVEN %s THEN returns given value', (input) => {
+					expect(ipPredicate.parse(input)).toBe(input);
+				});
+
+				test.each(['127.000.000.001', '127.0.0.1/24', 'fhqwhgads'])('GIVEN %s THEN throws a ConstraintError', (input) => {
+					expect(() => ipPredicate.parse(input)).toThrow(new ConstraintError('s.string.ip', 'Invalid ip address', input, 'expected.ip'));
```
vladfrangu, line 212: This message needs updating

tests/validators/string.test.ts
```diff
@@ -95,5 +95,154 @@ describe('StringValidator', () => {
...
+			describe('v4', () => {
+				const ipv4Predicate = s.string.ipv4;
+
+				test.each(['127.0.0.1'])('GIVEN %s THEN returns given value', (input) => {
+					expect(ipv4Predicate.parse(input)).toBe(input);
+				});
+
+				test.each(['127.000.000.001', '127.0.0.1/24', 'fhqwhgads'])('GIVEN %s THEN throws a ConstraintError', (input) => {
+					expect(() => ipv4Predicate.parse(input)).toThrow(
+						new ConstraintError('s.string.ipv4', 'Invalid ipv4 address', input, 'expected.ipv4')
```
vladfrangu, line 225: Ditto

tests/validators/string.test.ts
```diff
@@ -95,5 +95,154 @@ describe('StringValidator', () => {
...
+
+				test.each(['fhqwhgads', '2001:0db8:85a3:0000:0000:8a2e:0370:7334/24', '2001:0db8:85a3:0000:0000:8a2e:0370:7334/24/24'])(
+					'GIVEN %s THEN throws a ConstraintError',
+					(input) => {
+						expect(() => ipv6Predicate.parse(input)).toThrow(
+							new ConstraintError('s.string.ipv6', 'Invalid ipv6 address', input, 'expected.ipv6')
+						);
+					}
+				);
+			});
```
vladfrangu, line 245: Ditto

vladfrangu (changes requested):
Something for the future (not for you) is that we need to properly check error messages and contents. `toThrow` does not deeply check properties

tests/validators/string.test.ts
imranbarbhuiya: Should I add a new param in `stringRegex` to take the `expected` message? or rn there is no way to know which `UUID` version is used.

vladfrangu: As long as it doesn't slow it down too much, I'd say that would be ideal

imranbarbhuiya: updated

src/validators/StringValidator.ts
```diff
@@ -1,5 +1,18 @@
+import * as EmailValidator from 'email-validator';
+import { ConstraintError } from '..';
```
favna, line 2: Please remove all the `'..'` type imports and target the actual source files instead of the re-exports in `index.ts`

favna (changes requested):

src/constraints/util/emailValidator.ts
```diff
@@ -0,0 +1,64 @@
...
+	const account = emailParts[0];
+
+	// If the account name exceeds 64 characters then the email address is invalid
+	if (account.length > 64) return false;
+
+	// Extract the domain name of the email address
+	const domain = emailParts[1];
+
+	// If the domain name exceeds 255 characters then the email address is invalid
+	if (domain.length > 255) return false;
```
imranbarbhuiya, line 51: Can we do this length validation in the regex so we can avoid `exponential backtracking`.

favna: Feel free to adjust the regex but don't remove the if statements because string length checks are much faster than regex testing

imranbarbhuiya: I just realized string is getting passed to regex match after the if statements so there is no way of `exponential backtracking`. So this error isn't actually valid here. So are there anyways to ignore it?

kyranet: There is, there are RegExp expressions that it takes hours or even days even with 14 characters of input.

imranbarbhuiya: 🤔 What's the solution now?

imranbarbhuiya: 🤔 What's the solution now?

src/constraints/StringConstraints.ts
```diff
@@ -2,8 +2,22 @@ import { ConstraintError } from '../lib/errors/ConstraintError';
 import { Result } from '../lib/Result';
 import type { IConstraint } from './base/IConstraint';
 import { Comparator, eq, ge, gt, le, lt, ne } from './util/operators';
+import net from 'node:net';
```
kyranet, line 5: Please import the individual functions, don't import the default export.

src/constraints/StringConstraints.ts
```diff
@@ -44,3 +58,68 @@ export function stringLengthNe(length: number): IConstraint<string> {
...
+	};
+}
+
+export function stringUrl(options?: UrlOptions): IConstraint<string> {
+	return {
+		run(input: string) {
+			try {
+				const url = new URL(input);
+
+				// TODO: optimize theses checks
```
kyranet, line 79:
```suggestion
				// TODO(kyranet): optimize the option checks
```

src/constraints/StringConstraints.ts
```diff
@@ -44,3 +58,68 @@ export function stringLengthNe(length: number): IConstraint<string> {
 	const expected = `expected.length !== ${length}`;
 	return stringLengthComparator(ne, 's.string.lengthNe', expected, length);
 }
+
+export function stringRegex(regex: RegExp, type: 'url' | 'uuid' | 'regex', expected?: string): IConstraint<string> {
```
kyranet, line 62:
```suggestion
export function stringRegex(regex: RegExp, type: 'uuid' | 'regex', expected?: string): IConstraint<string> {
```

src/constraints/StringConstraints.ts
```diff
@@ -2,8 +2,22 @@ import { ConstraintError } from '../lib/errors/ConstraintError';
...
 import type { IConstraint } from './base/IConstraint';
 import { Comparator, eq, ge, gt, le, lt, ne } from './util/operators';
+import net from 'node:net';
 
-export type StringConstraintName = `s.string.length${'Lt' | 'Le' | 'Gt' | 'Ge' | 'Eq' | 'Ne'}`;
+export type StringConstraintName =
+	| `s.string.${`length${'Lt' | 'Le' | 'Gt' | 'Ge' | 'Eq' | 'Ne'}` | 'regex' | 'url' | 'uuid' | 'email'}`
+	| StringIpName;
+
+export type StringIpName = `s.string.ip${'v4' | 'v6' | ''}`;
```
kyranet, line 11: Shouldn't exist. Anything using this should instead cast to `StringConstraintName`.

src/validators/StringValidator.ts
```diff
@@ -29,6 +42,48 @@ export class StringValidator<T extends string> extends BaseValidator<T> {
...
 	}
 
+	public get email(): this {
+		return this.addConstraint({
+			run(input: string) {
+				return validateEmail(input)
+					? Result.ok(input)
+					: Result.err(new ConstraintError('s.string.email', 'Invalid email address', input, 'expected to be an email address'));
+			}
+		} as IConstraint<T>);
```
kyranet, line 52: Wait, why is this not with the other constraints?

tests/validators/string.test.ts
```diff
@@ -95,5 +95,159 @@ describe('StringValidator', () => {
...
+					);
+				});
+			});
+		});
+
+		describe('uuid', () => {
+			describe('default', () => {
+				const uuidPredicate = s.string.uuid();
+
+				test.each(['450d6a23-9e6f-45d9-9d5a-fd4f6e014f16', '00000000-0000-0000-0000-000000000000'])(
```
kyranet, line 159: UUIDs should move to variables `uuid4`, `uuid6`, and `uuidNull`, all of which should defined within `describe('uuid', cb)`.

tests/validators/string.test.ts
```diff
@@ -95,5 +95,159 @@ describe('StringValidator', () => {
...
+					new ConstraintError('s.string.regex', 'Invalid string format', input, `expected to match ${regex.source}`)
+				);
+			});
+		});
+
+		describe('ip', () => {
+			describe('default', () => {
+				const ipPredicate = s.string.ip();
+
+				test.each(['::1', '127.0.0.1'])('GIVEN %s THEN returns given value', (input) => {
```
kyranet, line 210: Same for IPs, they should be moved to variables.

kyranet (changes requested):

src/constraints/util/emailValidator.ts
kyranet: I'm leaning towards removing the domain part after `@` and then doing something like:
```js
new URL(`https://${domain}`);
```

The `validEmailRegex` would just validate the username, were it to need any testing.

However, the proposal above comes with some issues, we'd need to check for all of its properties to make sure `foo@hello.world/foo?hi=there` doesn't pass.

imranbarbhuiya:
I've changed `null` behavior in `UUID`. Now if u pass null it'll check for `NIL_UUID` as after checking the UUID package again I found it's not a versioned uuid it can be generated only with `NIL_UUID`

src/constraints/util/emailValidator.ts
imranbarbhuiya: done

imranbarbhuiya:
`new URL` allows urls like `https://ok`. Our email validator have a check for that but there are no checks present in urlValidator. It's a valid url so do we need to add a check?

src/constraints/StringConstraints.ts
```diff
@@ -44,3 +58,78 @@ export function stringLengthNe(length: number): IConstraint<string> {
...
+		run(input: string) {
+			return validateEmail(input)
+				? Result.ok(input)
+				: Result.err(new ConstraintError('s.string.email', 'Invalid email address', input, 'expected to be an email address'));
+		}
+	};
+}
+
+export function stringRegex(regex: RegExp, type: 'uuid' | 'regex', expected?: string): IConstraint<string> {
+	expected ??= `expected to match ${type === 'regex' ? regex.source : `a ${type}`}`;
```
kyranet, line 73: Come think of it, this method should be private under a separate name, and taking a RegExp, a `type` and an `expected`.

Then we do something like this, which helps making the code more maintainable and makes errors a lot faster, since no string interpolation happens every time an error is thrown:
```ts
export function stringRegex(regex: RegExp) {
	return stringRegexValidator('s.string.regex', `expected ${regex}.test(expected) to be true`, regex);
}

export interface StringUuidOptions {
	version?: 1 | 3 | 4 | 5 | `${bigint}-${bigint}` | null;
	nullable?: boolean;
}

export function stringUuid({ version = 4, nullable = false }: StringUuidOptions = {}) {
	version ??= '1-5';
	const regex = new RegExp(`^(?:[0-9A-F]{8}-[0-9A-F]{4}-[${version}][0-9A-F]{3}-[89AB][0-9A-F]{3}-[0-9A-F]{12}${nullable ? '|00000000-0000-0000-0000-000000000000' : ''})$`, 'i');
	const expected = `expected to match UUID${typeof version === 'number' ? `v${version}` : ` in range of ${version}`}`;
	return stringRegexValidator('s.string.uuid', expected, regex);
}
```

Obviously, the methods in the validator should use the options interface, and just pass it throught.

src/constraints/util/emailValidator.ts
```diff
@@ -0,0 +1,73 @@
...
+	if (domainParts.some((part) => part.length > 63)) return false;
+
+	// If all the checks above have passed then validate the entire email address against the email regex
+	return accountRegex.test(account) && validateEmailDomain(domain);
+}
+
+export function validateEmailDomain(domain: string): boolean {
+	try {
+		const url = new URL(`http://${domain}`);
+		return url.hostname === domain && url.pathname === '/' && !url.search;
```
kyranet, line 69: The hostname is enough, since that validates the entire domain with the `===`:
```suggestion
		return new URL(`http://${domain}`).hostname === domain;
```

src/constraints/StringConstraints.ts
kyranet: Oh yeah, and `NIL_UUID` should be removed, but we can probably do `version = 0` for exclusively null UUID, since it has a null version, wdyt @favna @vladfrangu

kyranet (changes requested):

src/constraints/StringConstraints.ts
```diff
@@ -44,3 +61,93 @@ export function stringLengthNe(length: number): IConstraint<string> {
...
+	return {
+		run(input: string) {
+			return validateEmail(input)
+				? Result.ok(input)
+				: Result.err(new ConstraintError('s.string.email', 'Invalid email address', input, 'expected to be an email address'));
+		}
+	};
+}
+
+export function stringRegexValidator(type: `s.string.${'uuid' | 'regex'}`, expected: string, regex: RegExp): IConstraint<string> {
```
kyranet, line 75: This is a private function, just like `stringLengthComparator`.
```suggestion
function stringRegexValidator(type: StringConstraintName, expected: string, regex: RegExp): IConstraint<string> {
```

tests/validators/string.test.ts
```diff
@@ -96,4 +96,208 @@ describe('StringValidator', () => {
...
+					);
+				});
+			});
+		});
+
+		describe('uuid', () => {
+			const uuid5s = ['2a7ff881-2944-55ae-94b0-b2ed34432297'];
+			const uuid4s = ['450d6a23-9e6f-45d9-9d5a-fd4f6e014f16'];
+			const uuid3s = ['2962d7f2-92f2-3105-8606-2234808bdfc8'];
+			const uuid1s = ['c310deb0-9785-11ec-8e65-592cb74aa664'];
```
kyranet, line 161: Those don't need to be arrays.

tests/validators/string.test.ts
```diff
@@ -96,4 +96,208 @@ describe('StringValidator', () => {
...
+			});
+
+			describe('uuid4', () => {
+				const uuid4Predicate = s.string.uuid({ version: 4 });
+
+				test.each(uuid4s)('GIVEN %s THEN returns given value', (input) => {
+					expect(uuid4Predicate.parse(input)).toBe(input);
+				});
+
+				test.each(['6e8bc430-9a1b-4f7f-b7a5', '6e8bc430-9a1b-4f7f-b7a5', '450d6a23-9e6f-35d9-9d5a-fd4f6e014f16'])(
```
kyranet, line 185: Why is this not using constants?

tests/validators/string.test.ts
```diff
@@ -96,4 +96,208 @@ describe('StringValidator', () => {
...
+					'GIVEN %s THEN throws a ConstraintError',
+					(input) => {
+						expect(() => uuid4Predicate.parse(input)).toThrow(
+							new ConstraintError('s.string.uuid', 'Invalid string format', input, 'expected UUID v4')
+						);
+					}
+				);
+
+				test('Given no version then match v4', () => {
```
kyranet, line 194: What's the meaning of this test?

tests/validators/string.test.ts
```diff
@@ -96,4 +96,208 @@ describe('StringValidator', () => {
...
+				});
+
+				test.each(uuid5s)('GIVEN %s THEN throws a ConstraintError', (input) => {
+					expect(() => uuidRangePredicate.parse(input)).toThrow(
+						new ConstraintError('s.string.uuid', 'Invalid string format', input, `expected UUID in range 1-4`)
+					);
+				});
+			});
+
+			test('Given NIL UUID THEN return NIL UUID', () => {
```
kyranet, line 229: This should be inside a `nullable` describe.

kyranet: Also don't name it `NIL UUID`, and the words `GIVEN` and `THEN` should be always uppercase. `return` should always be `returns`.

Syntax is as it follows: `GIVEN <given> THEN <returns|throws|rejects> <result>`.

kyranet (changes requested):

tests/validators/string.test.ts
imranbarbhuiya: https://github.com/uuidjs/uuid#uuidnil

It's name is `NIL` that's why I used `NIL`. So should I use `null` instead?

tests/validators/string.test.ts
imranbarbhuiya: If we don't specify a version then it'll check for uuid4.

src/constraints/StringConstraints.ts
```diff
@@ -44,3 +58,78 @@ export function stringLengthNe(length: number): IConstraint<string> {
 	const expected = `expected.length !== ${length}`;
 	return stringLengthComparator(ne, 's.string.lengthNe', expected, length);
 }
+
+export function stringEmail(): IConstraint<string> {
```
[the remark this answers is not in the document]
imranbarbhuiya: 🤔 then?

tests/validators/string.test.ts
kyranet: Yes, `NIL` is wording from other languages, here in JavaScript we use `null`.

tests/validators/string.test.ts
```diff
@@ -96,4 +96,215 @@ describe('StringValidator', () => {
...
+			});
+		});
+
+		describe('uuid', () => {
+			const uuid5 = '2a7ff881-2944-55ae-94b0-b2ed34432297';
+			const uuid4 = '450d6a23-9e6f-45d9-9d5a-fd4f6e014f16';
+			const uuid3 = '2962d7f2-92f2-3105-8606-2234808bdfc8';
+			const uuid1 = 'c310deb0-9785-11ec-8e65-592cb74aa664';
+			const nullUuid = '00000000-0000-0000-0000-000000000000';
+			const invalidUUids = ['6e8bc430-9a1b-4f7f-b7a5', '6e8bc430-9a1b-4f7f-b7a5'];
```
kyranet, line 163:
```suggestion
			const invalidUuids = ['6e8bc430-9a1b-4f7f-b7a5', '6e8bc430-9a1b-4f7f-b7a5'];
```

kyranet: And update all references to it.

kyranet (changes requested):
Last two things

src/constraints/StringConstraints.ts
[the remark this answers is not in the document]
kyranet: Actually, leave it. I plan on supporting email domain filters anyways.

kyranet (approved):

vladfrangu (changes requested):

kyranet (approved):

favna (approved):

vladfrangu (approved):

PR #39, opened 2022-02-24
feat: add tuple

imranbarbhuiya:
add tuple validator


src/validators/TupleValidator.ts
```diff
@@ -0,0 +1,41 @@
+import type { BaseError, IConstraint } from '..';
+import { BaseValidator } from './imports';
+import { Result } from '../lib/Result';
+import { ValidationError } from '../lib/errors/ValidationError';
+import { CombinedPropertyError } from '../lib/errors/CombinedPropertyError';
+
+export class TupleValidator<T extends BaseValidator<unknown>> extends BaseValidator<T[]> {
```
kyranet, line 7: This is entirely wrong, please read `UnionValidator`'s type.

src/validators/TupleValidator.ts
```diff
@@ -0,0 +1,41 @@
...
+		this.validators = validators;
+	}
+
+	protected override clone(): this {
+		return Reflect.construct(this.constructor, [this.validators, this.constraints]);
+	}
+
+	protected handle(values: unknown): Result<T[], ValidationError | CombinedPropertyError> {
+		if (!Array.isArray(values)) {
+			return Result.err(new ValidationError('TupleValidator', 'Expected a tuple', values));
```
kyranet, line 21:
```suggestion
			return Result.err(new ValidationError('TupleValidator', 'Expected an array', values));
```

src/validators/TupleValidator.ts
```diff
@@ -0,0 +1,41 @@
...
+		}
+
+		if (values.length !== this.validators.length) {
+			return Result.err(new ValidationError('TupleValidator', `Expected a tuple of length ${this.validators.length}`, values));
```
kyranet, line 25:
```suggestion
			return Result.err(new ValidationError('TupleValidator', `Expected an array of length ${this.validators.length}`, values));
```

kyranet (changes requested):

src/validators/TupleValidator.ts
imranbarbhuiya: `UnionValidator`'s type is also wrong. Try to use it. 

1. It only allows rest params not an array.
2. It allows only 1 type of `BaseValidator` means u can use `union([s.string, s.number])`

I've sent a ss in discord server. Please check

favna: @kyranet what's the latest on this?

src/validators/TupleValidator.ts
```diff
@@ -0,0 +1,41 @@
+import type { BaseError, IConstraint } from '..';
```
favna, line 1: Please remove all the `'..'` type imports and target the actual source files instead of the re-exports in `index.ts`

favna (dismissed):

src/validators/TupleValidator.ts
imranbarbhuiya: Vsc auto import. I'll fix it tomorrow

favna: Figured as much yeah

src/validators/TupleValidator.ts
kyranet: I'll check it when I have time.

src/lib/Shapes.ts
```diff
@@ -101,3 +106,7 @@ export class Shapes {
 		return new MapValidator(keyValidator, valueValidator);
 	}
 }
+
+type UnwrapTuple<T extends [...any[]]> = T extends [infer Head, ...infer Tail]
```
vladfrangu, line 110: Shouldn't we export this? Might be useful for some people 👀

vladfrangu (approved):

src/lib/Shapes.ts
kyranet: Why... would you do that? I mean, it's an *extremely specific* recursive type unwrapper for Shapeshift

favna (approved):

PR #41, opened 2022-02-25
fix: fix union type and add test

imranbarbhuiya:
I've tried to fix union type but if this isn't the best way of doing this then lmk how I can do it.

### DIFF
--- a/src/lib/Shapes.ts
+++ b/src/lib/Shapes.ts
@@ -6,16 +6,16 @@ import {
 	DateValidator,
 	InstanceValidator,
 	LiteralValidator,
+	MapValidator,
 	NeverValidator,
 	NullishValidator,
 	NumberValidator,
 	ObjectValidator,
 	PassthroughValidator,
+	RecordValidator,
 	SetValidator,
 	StringValidator,
-	UnionValidator,
-	RecordValidator,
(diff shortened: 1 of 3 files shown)


src/lib/Shapes.ts
```diff
@@ -81,8 +81,8 @@ export class Shapes {
 		return new InstanceValidator(expected);
 	}
 
-	public union<T>(...validators: readonly BaseValidator<T>[]) {
-		return new UnionValidator(validators);
+	public union<T extends BaseValidator<unknown>>(validators: readonly T[]): T {
```
kyranet, line 84: This return type is wrong, it doesn't return T, it returns an UnionValidator with the unwrapped type.

Also the rest params are intentional, if you use an array, just spread it.

kyranet (changes requested):

kyranet (dismissed):

src/lib/Shapes.ts
imranbarbhuiya: Ok then readme needs a fix.

vladfrangu (approved):

kyranet (approved):

favna (approved):

PR #44, opened 2022-02-26
test: add more tests

imranbarbhuiya:


imranbarbhuiya:
All the overridden clone method's tests are missing. These are protected methods so without adding `ignore` comment can't add test. How can I add them?

imranbarbhuiya:
I've no clue about the remaining tests. Lmk when these lines will be used.

imranbarbhuiya:
97% coverage done 🎉 🎉

tests/validators/array.test.ts
```diff
@@ -111,4 +123,23 @@ describe('ArrayValidator', () => {
 			});
 		});
 	});
+
+	describe('Clone', () => {
+		// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+		// @ts-expect-error
```
favna, line 129:
```suggestion
		// @ts-expect-error Accessing protected method
```

I think? Anyway if you add a reason then you don't need the eslint ignore.

tests/validators/number.test.ts
```diff
@@ -276,5 +276,14 @@ describe('NumberValidator', () => {
 				expect(defaultFunctionPredicate.parse(undefined)).toBe(5);
 			});
 		});
+
+		describe('Clone', () => {
+			// @ts-ignore Test clone
```
favna, line 281:
```suggestion
			// @ts-expect-error Test clone
```

tests/validators/map.test.ts
```diff
@@ -32,4 +32,11 @@ describe('MapValidator', () => {
 			])
 		);
 	});
+
+	test('Clone', () => {
+		// @ts-ignore Test clone
```
favna, line 37:
```suggestion
		// @ts-expect-error Test clone
```

tests/validators/instance.test.ts
```diff
@@ -13,4 +13,11 @@ describe('InstanceValidator', () => {
 	test("GIVEN anything which isn't and instance of User THEN throws ValidationError", () => {
 		expect(() => predicate.parse(123)).toThrow(new ExpectedValidationError('InstanceValidator', 'Expected', 123, Array));
 	});
+
+	test('Clone', () => {
+		// @ts-ignore Test clone
```
favna, line 18:
```suggestion
		// @ts-expect-error Test clone
```

tests/validators/literal.test.ts
```diff
@@ -0,0 +1,22 @@
...
+	});
+
+	test("GIVEN anything which isn't the literal THEN throws ExpectedValidationError", () => {
+		expect(() => predicate.parse('hello')).toThrow(
+			new ExpectedValidationError('LiteralValidator', 'Expected values to be equals', 'hello', 'sapphire')
+		);
+	});
+
+	test('Clone', () => {
+		// @ts-ignore Test clone
```
favna, line 17:
```suggestion
		// @ts-expect-error Test clone
```

tests/validators/record.test.ts
```diff
@@ -24,4 +24,11 @@ describe('RecordValidator', () => {
 			])
 		);
 	});
+
+	test('Clone', () => {
+		// @ts-ignore Test clone
```
favna, line 29:
```suggestion
		// @ts-expect-error Test clone
```

tests/validators/set.test.ts
```diff
@@ -18,4 +18,17 @@ describe('SetValidator', () => {
...
 	});
+
+	test('s.string.set and s.set(s.string) should return same', () => {
+		const set = new Set(['hello']);
+
+		expect(predicate.parse(set)).toStrictEqual(s.string.set.parse(set));
+	});
+
+	test('Clone', () => {
+		// @ts-ignore Test clone
```
favna, line 29:
```suggestion
		// @ts-expect-error Test clone
```

tests/validators/union.test.ts
```diff
@@ -22,4 +22,67 @@ describe('UnionValidator', () => {
...
+		expect(() => nullishPredicate.parse(false)).toThrow(
+			new CombinedError([
+				new ValidationError('s.string', 'Expected a string primitive', false),
+				new ValidationError('s.number', 'Expected a number primitive', false)
+			])
+		);
+	});
+
+	test('Clone', () => {
+		// @ts-ignore Test clone
```
favna, line 82:
```suggestion
		// @ts-expect-error Test clone
```

favna (changes requested):
`ts-expect-error` achieves the same as `ts-ignore` but is far better.
Docs: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-9.html#-ts-expect-error-comments

favna (dismissed):

tests/validators/date.test.ts
```diff
@@ -70,6 +70,10 @@ describe('DateValidator', () => {
 		test.each([new Date('2021-01-02'), new Date('2023-01-02')])('GIVEN %s THEN throws ConstraintError', (value) => {
 			expect(() => eqPredicate.parse(value)).toThrow(new ConstraintError('s.date.eq', 'Invalid Date value', value, 'expected = 2022-01-02'));
 		});
+
+		test.each(['abc', '', null, undefined])('GIVEN %s THEN throws ConstraintError', (value) => {
+			expect(() => eqPredicate.parse(value)).toThrow(new ConstraintError('s.date.ge', 'Expected a Date', value, 'expected = 2022-01-02'));
```
kyranet, line 75: This feels wrong, it shouldn't throw a ConstraintError, but a ValidationError.

Jest, why does your `.toThrow` suck so much?

tests/validators/passthrough.test.ts
```diff
@@ -0,0 +1,17 @@
...
+
+describe('AnyValidator', () => {
+	const predicate = s.any;
+
+	test.each([1, 'hello', null])('GIVEN anything %s THEN returns the given value', (input) => {
+		expect<any>(predicate.parse(input)).toBe(input);
+	});
+});
+
+describe('unknownValidator', () => {
```
kyranet, line 11:
```suggestion
describe('UnknownValidator', () => {
```

tests/validators/date.test.ts
kyranet: Also, the `test.each` should be moved to line 11, which is where it tests for invalid input.

tests/validators/date.test.ts
```diff
@@ -82,5 +86,9 @@ describe('DateValidator', () => {
 		test.each([new Date('2022-01-02')])('GIVEN %s THEN throws ConstraintError', (value) => {
 			expect(() => nePredicate.parse(value)).toThrow(new ConstraintError('s.date.ne', 'Invalid Date value', value, 'expected != 2022-01-02'));
 		});
+
+		test.each(['abc', '', null, undefined])('GIVEN %s THEN throws ConstraintError', (value) => {
+			expect(() => nePredicate.parse(value)).toThrow(new ConstraintError('s.date.ne', 'Expected a Date', value, 'expected != 2022-01-02'));
+		});
```
kyranet, line 92: See above:
```suggestion
```

tests/validators/array.test.ts
```diff
@@ -11,6 +11,18 @@ describe('ArrayValidator', () => {
 		expect(() => predicate.parse('Hello there')).toThrow(new ValidationError('ArrayValidator', 'Expected an array', 'Hello there'));
 	});
 
+	const invalidArray = [[123], [true], [{}], [[]], [null]];
+
+	test.each([invalidArray])('Given an array with value %s other than string THEN throws CombinedPropertyError', (input) => {
+		expect(() => predicate.parse(input)).toThrow(
```
kyranet, line 17:
```suggestion
	const invalidArray = [123, true, {}, [], null];

	test.each(invalidArray)('Given an array with value %s other than string THEN throws CombinedPropertyError', (input) => {
		expect(() => predicate.parse(input)).toThrow(
```

tests/validators/array.test.ts
```diff
@@ -111,4 +123,22 @@ describe('ArrayValidator', () => {
...
+		test.each([
+			[
+				['Hello', 'there'],
+				['Hello', 'there']
+			],
+			[
+				['Hello', 'there', 'foo'],
+				['Hello', 'there', 'foo']
+			]
+		])('GIVEN %p THEN returns given value', (value) => {
```
kyranet, line 140: We don't need to define 2 arguments.
```suggestion
		test.each([
			[['Hello', 'there']],
			[['Hello', 'there', 'foo']]
		])('GIVEN %p THEN returns given value', (value) => {
```

tests/validators/enum.test.ts
```diff
@@ -0,0 +1,19 @@
+import { CombinedError, ExpectedValidationError, s } from '../../src';
+
+describe('EnumValidator', () => {
+	const predicate = s.enum('a', 'b', 'c');
+
+	test.each(['a', 'b', 'c'])('GIVEN a string %s THEN returns a string', (input) => {
+		expect(predicate.parse(input)).toBe(input);
+	});
+
+	test.each(['d', 'e', 'f'])('GIVEN a string %s THEN throws CombinedError', (input) => {
```
kyranet, line 10: Should test for invalid values as well, not just strings.

tests/validators/literal.test.ts
```diff
@@ -0,0 +1,22 @@
+import { ExpectedValidationError, s } from '../../src';
+
+describe('LiteralValidator', () => {
```
kyranet, line 3: Missing test for Date's special case:

https://github.com/sapphiredev/shapeshift/blob/dc7ccaf9f2bac0513fbb58f5ac0676e2fc2b83e2/src/lib/Shapes.ts#L76

tests/validators/null.test.ts
```diff
@@ -0,0 +1,9 @@
+import { s } from '../../src';
+
+describe('null', () => {
```
kyranet, line 3:
```suggestion
describe('NullValidator', () => {
```

tests/validators/number.test.ts
```diff
@@ -276,5 +276,14 @@ describe('NumberValidator', () => {
 				expect(defaultFunctionPredicate.parse(undefined)).toBe(5);
 			});
 		});
+
+		describe('Clone', () => {
+			// @ts-expect-error Test clone
+			const clonePredicate = s.number.default(5).clone();
+
+			test('GIVEN undefined THEN returns the default value', () => {
+				expect(clonePredicate.parse(undefined)).toBe(5);
```
kyranet, line 285: I don't get it, why don't we just strict equals the instances?

tests/validators/record.test.ts
```diff
@@ -24,4 +24,11 @@ describe('RecordValidator', () => {
 			])
 		);
 	});
+
+	test('Clone', () => {
```
kyranet, line 28:
```suggestion
	test('GIVEN clone THEN returns similar instance', () => {
```

kyranet: Applies to everything else that's a `test()` as well.

tests/validators/string.test.ts
```diff
@@ -96,4 +96,44 @@ describe('StringValidator', () => {
 			});
 		});
 	});
+
+	describe('Nullable', () => {
+		const nullablePredicate = s.string.nullable;
```
kyranet, line 101: Isn't this a base validator? Why is it in string?

tests/validators/string.test.ts
```diff
@@ -96,4 +96,44 @@ describe('StringValidator', () => {
...
+			expect<string | null>(nullablePredicate.parse(null)).toBe(null);
+		});
+
+		test('GIVEN a string THEN returns a string', () => {
+			expect(nullablePredicate.parse('Hello There')).toBe('Hello There');
+		});
+	});
+
+	describe('Nullish', () => {
+		const nullishPredicate = s.string.nullish;
```
kyranet, line 113: Same here (base validator methods)

tests/validators/string.test.ts
```diff
@@ -96,4 +96,44 @@ describe('StringValidator', () => {
...
+			expect(nullishPredicate.parse(undefined)).toBe(undefined);
+		});
+
+		test('GIVEN a string THEN returns a string', () => {
+			expect(nullishPredicate.parse('Hello There')).toBe('Hello There');
+		});
+	});
+
+	describe('Or', () => {
+		const orPredicate = s.string.or(s.number);
```
kyranet, line 129: Same here (base validator methods)

kyranet: Also `s.string.or(s.number)` should validate for whether it's strict equals to `s.union(s.string, s.number)`.

tests/validators/undefined.test.ts
```diff
@@ -0,0 +1,9 @@
+import { s } from '../../src';
+
+describe('undefined', () => {
```
kyranet, line 3:
```suggestion
describe('UndefinedValidator', () => {
```

tests/validators/undefined.test.ts
```diff
@@ -0,0 +1,9 @@
...
+	const predicate = s.undefined;
+
+	test('GIVEN undefined should return undefined', () => {
```
kyranet, line 6: No shoulds, must.
```suggestion
	test('GIVEN undefined THEN returns undefined', () => {
```

kyranet (changes requested):

tests/validators/date.test.ts
kyranet: It does, but `.toThrow` literally only tests for `error.message`, doesn't check anything else.

The fact this passes **WITHOUT** instanceof checks is beyond my understanding,

favna: That's what they call design decisions. Nothing wrong there. Simply add a `.toBeInstanceOf` assertion before `.toThrow` and it's done. 

https://jestjs.io/docs/expect#tobeinstanceofclass

kyranet: Doesn't remove the fact it doesn't check the other properties. I'm leaning towards making our own macros calling `expect`.

favna: Regardless, this is a GitHub review comment. Not a place for you to vent and rant. Keep the review comments objective and useful please.

tests/validators/array.test.ts
imranbarbhuiya: Here I want to pass `[123]` and the array is an array of string so it should throw combinedError.

Also If I just pass `invalidArray` then it'll pass `123`. If I keep the array as 

```js
const invalidArray = [[123], [true], [{}], [[]], [null]];
```

and pass 
```js
test.each(invalidArray)
```
then also jest makes the array 1D. Means it's still passing `123`. So this one is correct.

ref: https://jestjs.io/docs/api#testeachtablename-fn-timeout

tests/validators/literal.test.ts
```diff
@@ -0,0 +1,32 @@
...
+		expect(() => predicate.parse('hello')).toThrow(
+			new ExpectedValidationError('LiteralValidator', 'Expected values to be equals', 'hello', 'sapphire')
+		);
+	});
+
+	test('GIVEN clone THEN returns similar instance', () => {
+		// @ts-expect-error Test clone
+		const clonePredicate = predicate.clone();
+
+		expect(clonePredicate).toBeInstanceOf(predicate.constructor);
```
imranbarbhuiya, line 20: @kyranet did u mean something like this?

tests/validators/array.test.ts
kyranet: I'm very well aware of table arguments, they're just very hard to read properly, not to mention also inconsistent with the rest of the tests.

tests/validators/literal.test.ts
kyranet: Instanceof + strictEquals, so all properties are validated

tests/validators/base.test.ts
```diff
@@ -0,0 +1,31 @@
+import { s } from '../../src';
+
+describe('BaseValidator', () => {
```
kyranet, line 3: Missing `.optional`, `.array`, `.set`, `.transform`, and `.default`.

tests/validators/literal.test.ts
```diff
@@ -13,10 +13,20 @@ describe('LiteralValidator', () => {
...
+	});
+
+	describe('DateLiteral', () => {
+		const date = new Date('2022-01-01');
+		const dateLiteralPredicate = s.literal(date);
+		const datePredicate = s.date;
+
+		test('GIVEN a date literal THEN returns the given value', () => {
+			expect(dateLiteralPredicate.parse(date)).toStrictEqual(datePredicate.parse(date));
```
kyranet, line 29: Should test for instances with the same value, not just the instance used in the literal.

tests/validators/union.test.ts
```diff
@@ -78,7 +78,15 @@ describe('UnionValidator', () => {
...
 	});
 
-	test('Clone', () => {
+	describe('Or', () => {
+		const orPredicate = s.string.or(s.number);
+
+		test('Or and union should be strict equal', () => {
+			expect(unionPredicate).toStrictEqual(orPredicate);
+		});
+	});
```
kyranet: This is not testing for the `.or` override in UnionValidator, it's testing BaseValidator's `.or`.

tests/validators/base.test.ts
```diff
@@ -0,0 +1,43 @@
...
+	describe('Nullable', () => {
```
kyranet, line 4: If a describe references a method name, it should be named the same as the method, including casing.

tests/validators/date.test.ts
```diff
@@ -8,79 +8,93 @@ describe('DateValidator', () => {
...
-	describe('lt', () => {
-		const ltPredicate = s.date.lt(new Date('2022-02-01'));
+	describe('Comparator', () => {
+		describe('lt', () => {
+			const ltPredicate = s.date.lt(new Date('2022-02-01'));
 
-		test.each([new Date('2022-01-01'), new Date('2020-01-01')])('GIVEN %s THEN returns given value', (value) => {
-			expect(ltPredicate.parse(value)).toBe(value);
-		});
+			test.each([new Date('2022-01-01'), new Date('2020-01-01')])('GIVEN %s THEN returns given value', (value) => {
```
kyranet, line 19: The dates used in `test.each` should be moved to variables.

kyranet (changes requested):

tests/validators/union.test.ts
imranbarbhuiya: Where you've found this line? It's from a old commit ig

tests/validators/array.test.ts
imranbarbhuiya: Your changes aren't passing the test. It won't throw a `CombinedError` instead it'll throw a `ValidationError` and `ValidationError` is added just above this test.

tests/validators/union.test.ts
```diff
@@ -22,4 +22,75 @@ describe('UnionValidator', () => {
 			])
 		);
 	});
+
+	describe('or', () => {
+		test('Union Or', () => {
```
kyranet, line 27: All tests must use `GIVEN ... THEN ...`

kyranet: For validations, we always want to use at least 1 passing and at least 1 failing test, so those should use `describe`, like the optional test in this file.

kyranet (changes requested):

tests/validators/union.test.ts
imranbarbhuiya: I'm bad at wording can u please update this line?

tests/validators/undefined.test.ts
```diff
@@ -0,0 +1,9 @@
+import { s } from '../../src';
+
+describe('UndefinedValidator', () => {
```
kyranet, line 3: Missing tests for non-undefined

tests/validators/base.test.ts
```diff
@@ -0,0 +1,90 @@
+import { s } from '../../src';
+
+describe('BaseValidator', () => {
+	describe('#optional', () => {
```
kyranet, line 4:
```suggestion
	describe('optional', () => {
```

kyranet: And same for the others

kyranet (changes requested):

tests/validators/literal.test.ts
imranbarbhuiya: you mean `toStrictEqual(date)`?

kyranet: As a matter of a fact, I highly doubt Jest checks date equality properly, you may need `.getTime()`.

But no, I'm referring that you should construct a new instance with the exact same time, and see that it matches.

tests/validators/base.test.ts
```diff
@@ -0,0 +1,90 @@
+import { s } from '../../src';
+
+describe('BaseValidator', () => {
+	describe('optional', () => {
```
kyranet, line 4: Should test non-optional (null) and other invalid types.

tests/validators/base.test.ts
```diff
@@ -0,0 +1,90 @@
...
+		const optionalPredicate = s.string.optional;
+
+		test.each([undefined, 'hello'])('GIVEN %s THEN returns given value', (input) => {
+			expect(optionalPredicate.parse(input)).toEqual(input);
+		});
+	});
+
+	describe('nullable', () => {
```
kyranet, line 12: Should test non-nullable (undefined) and other invalid types.

tests/validators/base.test.ts
```diff
@@ -0,0 +1,90 @@
...
+
+	describe('nullish', () => {
+		const nullishPredicate = s.string.nullish;
+
+		test.each(['Hello There', undefined, null])('GIVEN %s THEN returns the given value', (input) => {
+			expect<string | undefined | null>(nullishPredicate.parse(input)).toBe(input);
+		});
+	});
+
+	describe('array', () => {
```
kyranet, line 28: Should test it's identical to `s.array(s.number)`

tests/validators/base.test.ts
```diff
@@ -0,0 +1,90 @@
...
+		const arrayPredicate = s.number.array;
+
+		test('GIVEN an array of string THEN returns the given value', () => {
+			expect(arrayPredicate.parse([1, 2, 3])).toStrictEqual([1, 2, 3]);
+		});
+	});
+
+	describe('set', () => {
```
kyranet, line 36: Should test it's identical to `s.set(s.number)`

tests/validators/base.test.ts
```diff
@@ -0,0 +1,90 @@
...
+		const setPredicate = s.number.set;
+
+		test('GIVEN a set of string THEN returns the given value', () => {
+			expect(setPredicate.parse(new Set([1, 2, 3]))).toStrictEqual(new Set([1, 2, 3]));
+		});
+	});
+
+	describe('or', () => {
```
kyranet, line 44: Should test invalid types as well.

tests/validators/base.test.ts
```diff
@@ -0,0 +1,90 @@
...
+		const orPredicate = s.string.or(s.number);
+
+		test.each(['Hello There', 6])('GIVEN a string or number THEN returns a string', (input) => {
+			expect(orPredicate.parse(input)).toBe(input);
+		});
+	});
+
+	describe('transform', () => {
```
kyranet, line 52: We probably want to test transforms inside an union, e.g. what we have and `.or(s.number)`, to see that it only calls the callback when it's a string.

That'd be another test.

tests/validators/set.test.ts
```diff
@@ -18,4 +18,18 @@ describe('SetValidator', () => {
 
 		expect(() => predicate.parse(set)).toThrow(new CombinedError([new ValidationError('StringValidator', 'Expected a string', input)]));
 	});
+
+	test('s.string.set and s.set(s.string) should return same', () => {
```
kyranet, line 22:
```suggestion
	test('GIVEN s.string.set THEN returns s.set(s.string)', () => {
```

kyranet: But this is for the base validator, it shouldn't be here.

tests/validators/base.test.ts
```diff
@@ -0,0 +1,90 @@
+import { s } from '../../src';
+
+describe('BaseValidator', () => {
```
kyranet, line 3: Missing tests to verify that all methods and getters return a clone and not the instance itself.

kyranet: Also the tests should use the generic to verify the return types are correct.

kyranet (changes requested):
Very close now!

tests/validators/base.test.ts
imranbarbhuiya: I've added a test. Check if this is how u want?

kyranet (dismissed):

tests/validators/array.test.ts
```diff
@@ -111,4 +119,13 @@ describe('ArrayValidator', () => {
 			});
 		});
 	});
+
+	test('GIVEN clone THEN returns similar instance', () => {
+		const arrayPredicate = s.string.array;
+		// @ts-expect-error Test clone
```
vladfrangu, line 125: Why is clone require-ing a ts-expect-error 👀

tests/validators/base.test.ts
```diff
@@ -0,0 +1,231 @@
...
+
+			expect(orUnionPredicate).toBeInstanceOf(unionPredicate.constructor);
+			expect(unionPredicate).toStrictEqual(orUnionPredicate);
+		});
+	});
+
+	describe('transform', () => {
+		const transformPredicate = s.string.transform((value) => value.toUpperCase());
+
+		test('GIVEN a string THEN returns a number', () => {
```
vladfrangu, line 110:
```suggestion
		test('GIVEN a string THEN returns uppercase string', () => {
```

tests/validators/base.test.ts
```diff
@@ -0,0 +1,231 @@
...
+		const unionTransformPredicate = s.string.transform((value) => value.toUpperCase()).or(s.number);
+
+		test('GIVEN string THEN returns uppercase string', () => {
+			expect<string | number>(unionTransformPredicate.parse('Hello There')).toStrictEqual('HELLO THERE');
+		});
+		test('GIVEN number THEN returns number', () => {
+			expect<string | number>(unionTransformPredicate.parse(6)).toStrictEqual(6);
+		});
+
+		test.each([false, true, null])('GIVEN %s THEN throws CombinedError', (input) => {
```
vladfrangu, line 123:
```suggestion
		test.each([false, true, null, undefined])('GIVEN %s THEN throws CombinedError', (input) => {
```

tests/validators/null.test.ts
```diff
@@ -0,0 +1,13 @@
+import { ExpectedValidationError, s } from '../../src';
+
+describe('NullValidator', () => {
+	const predicate = s.null;
+
+	test('GIVEN null THEN returns null', () => {
+		expect(predicate.parse(null)).toBe(null);
+	});
+
+	test.each([undefined, 123, 'Hello'])('GIVEN non-null %s THEN throws ExpectedValidationError', (input) => {
```
vladfrangu, line 10:
```suggestion
	test.each([undefined, 123, 'Hello', {}])('GIVEN non-null %s THEN throws ExpectedValidationError', (input) => {
```

vladfrangu (changes requested):

tests/validators/array.test.ts
imranbarbhuiya: It's a protected method

imranbarbhuiya:
@kyranet can u rerun the test?

vladfrangu (approved):

favna (approved):

kyranet (approved):

favna:
@all-contributors please add @imranbarbhuiya for code and test and bug and doc

PR #53, opened 2022-03-02
feat(ArrayValidator): add length ranges

vladfrangu:


src/constraints/ArrayLengthConstraints.ts
```diff
@@ -44,3 +44,25 @@ export function arrayLengthNe<T>(value: number): IConstraint<T[]> {
...
+	return {
+		run(input: T[]) {
+			return input.length >= start && input.length < endBefore //
+				? Result.ok(input)
+				: Result.err(new ConstraintError('s.array(T).lengthRange', 'Invalid Array length', input, expected));
+		}
+	};
+}
+
+export function arrayLengthRangeInclusive<T>(start: number, endBefore: number): IConstraint<T[]> {
```
kyranet, line 59: `endBefore` should be named `end`.

kyranet (changes requested):

README.md
```diff
@@ -210,6 +210,8 @@ s.string.array.lengthGt(5); // Must have more than 5 elements
 s.string.array.lengthGe(5); // Must have 5 or more elements
 s.string.array.lengthEq(5); // Must have exactly 5 elements
 s.string.array.lengthNe(5); // Must not have exactly 5 elements
+s.string.array.lengthRange(0, 4); // Must have at least 0 elements and less than 4 elements (in math, that is [0, 4))
```
favna, line 213: Missing closing ] here

favna, line 213: Between this and inclusive it isn't really made clear what the difference is without taking the actual function names into account because this says [0, 4] and inclusive says [0, 5]. Shouldn't this be [0, 3] or [1, 3]? 

And if not [1, 3] (when giving it 0,4 as ranges) then please also add rangeExclusive which specifically does everything except the boundaries in both sides.

In general I recommend using the same ranges for each example so the difference is clearly shown. That is to say, it should be clear from just the documentation what will and won't be included given specific ranges and functions. One idea is to just write out a full example using a simple array of `[1, 2, 3, 4]` then listing when it will give a validation error for a specific function and when it will not.

favna (changes requested):

README.md
imranbarbhuiya: If I'm not wrong then `[0,4)` means including 0 but not 4.

imranbarbhuiya: rangeInclusive includes 4 so `[0,4]` means it can be of length `0, 1, 2, 3, 4` means it can have length of 4. But for range `[0, 4)` the range will be `0, 1, 2, 3` not including 4.

vladfrangu: I am not missing a closing `]` at all! Math range of `[0, 4)` means 0, 1, 2, 3

vladfrangu: > please also add rangeExclusive which specifically does everything except the boundaries in both sides

Done!

kyranet (dismissed):

favna (approved):

kyranet (approved):

PR #61, opened 2022-03-09
perf(s.string): speed up URL and IP constraints

kyranet:
Resolving a TODO and then improving IP's error reporting performance

### DIFF
--- a/src/constraints/util/urlValidators.ts
+++ b/src/constraints/util/urlValidators.ts
@@ -1,4 +1,4 @@
-import { ConstraintError } from '../../lib/errors/ConstraintError';
+import { MultiplePossibilitiesConstraintError } from '../../lib/errors/MultiplePossibilitiesConstraintError';
 import { combinedErrorFn, ErrorFn } from './common/combinedResultFn';
 
 export type StringProtocol = `${string}:`;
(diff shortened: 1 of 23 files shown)


src/constraints/util/urlValidators.ts
```diff
@@ -0,0 +1,36 @@
...
+function allowedProtocolsFn(allowedProtocols: StringProtocol[]) {
+	const oneOf = allowedProtocols.join(', ');
+	return (input: string, url: URL) =>
+		allowedProtocols.includes(url.protocol as StringProtocol)
+			? null
+			: new ConstraintError('s.string.url', 'Invalid URL protocol', input, `expected ${url.protocol} to be one of: ${oneOf}`);
+}
+
+function allowedDomainsFn(allowedDomains: StringDomain[]) {
+	const oneOf = allowedDomains.join(', ');
```
vladfrangu, line 31: Can you move this in the error definition instead, as it's a useless join call for valid domains

src/constraints/util/urlValidators.ts
```diff
@@ -0,0 +1,36 @@
...
+	const fns: ErrorFn<[input: string, url: URL], ConstraintError<string>>[] = [];
+
+	if (options?.allowedProtocols?.length) fns.push(allowedProtocolsFn(options.allowedProtocols));
+	if (options?.allowedDomains?.length) fns.push(allowedDomainsFn(options.allowedDomains));
+
+	return combinedErrorFn(...fns);
+}
+
+function allowedProtocolsFn(allowedProtocols: StringProtocol[]) {
+	const oneOf = allowedProtocols.join(', ');
```
vladfrangu, line 23: Ditto

vladfrangu (changes requested):

src/constraints/util/urlValidators.ts
kyranet: ... what

vladfrangu: You take this join call
You remove it from here
You move it in the `new Error(` part where its used
You profit

kyranet: Mate, that's not executed in valid domains.

vladfrangu: Useless execution still :)

vladfrangu (approved):

favna (approved):

PR #62, opened 2022-03-12
fix(s.union): fix union overrides

kyranet:
The overridden methods were not supposed to nest UnionValidators

### DIFF
--- a/src/validators/UnionValidator.ts
+++ b/src/validators/UnionValidator.ts
@@ -33,7 +33,7 @@ export class UnionValidator<T> extends BaseValidator<T> {
 			return this.clone();
 		}
 
-		return new UnionValidator([new LiteralValidator(undefined), this.clone()]);
+		return new UnionValidator([new LiteralValidator(undefined), ...this.validators]);
 	}
 
 	public override get nullable(): UnionValidator<T | null> {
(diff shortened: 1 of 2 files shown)


src/validators/UnionValidator.ts
```diff
@@ -56,19 +56,24 @@ export class UnionValidator<T> extends BaseValidator<T> {
...
 	public override get nullish(): UnionValidator<T | null | undefined> {
 		if (this.validators.length === 0) return new UnionValidator<T | null | undefined>([new NullishValidator()], this.constraints);
 
 		const [validator] = this.validators;
-		if (validator instanceof NullishValidator) {
+		if (validator instanceof LiteralValidator) {
+			// If already nullable or optional, promote the union to nullish:
+			if (validator.expected === null || validator.expected === undefined) {
+				return new UnionValidator<T | null | undefined>([new NullishValidator(), ...this.validators.slice(1)], this.constraints);
+			}
```
vladfrangu, line 70: Handle only undefined and only null too? Because this will changes the behavior of s.union(s.literal(null), s.string) from string | null to string | undefined | null

vladfrangu (changes requested):

vladfrangu (approved):

favna (approved):

PR #70, opened 2022-03-13
test: increase test coverage for email validation

favna:
TODO:
- [x] The statement `if (email.length - lastDotIndex > 63) return false;` is currently uncovered and I also have no idea what input to give to cover it. Please advice.
- [x] Giving `short-account-name@domain-name.-that-has-more-than-sixty-three-characters.com` as input throws a `ReferenceError`, unknown why. This should be solved.

### DIFF
--- a/tests/common/macros/comparators.ts
+++ b/tests/common/macros/comparators.ts
@@ -29,7 +29,7 @@ export function expectModifiedClonedValidator<T>(expected: BaseValidator<T>, act
 export function expectError<T = any>(cb: () => T, expected: BaseError) {
 	try {
 		cb();
-		fail('expected to throw, but failed to do so');
+		fail();
 	} catch (error) {
 		expect(error).toBeDefined();
 		expectIdenticalError(error as BaseError, expected);
--- a/tests/common/macros/comparators.ts
+++ b/tests/common/macros/comparators.ts
@@ -29,12 +29,13 @@ export function expectModifiedClonedValidator<T>(expected: BaseValidator<T>, act
 export function expectError<T = any>(cb: () => T, expected: BaseError) {
 	try {
 		cb();
-		fail();
 	} catch (error) {
 		expect(error).toBeDefined();
 		expectIdenticalError(error as BaseError, expected);
-		// expect(removeStack(error as Error)).toStrictEqual(removeStack(expected));
+		return;
 	}
+
+	fail('Expected to throw, but failed to do so');
 }
 
 function expectIdenticalError(actual: BaseError, expected: BaseError) {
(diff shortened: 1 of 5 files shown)


tests/jest.setup.ts
```diff
@@ -0,0 +1,5 @@
+function fail(reason = 'Expected to throw, but failed to do so') {
+	throw new Error(reason);
+}
+
+global.fail = fail;
```
kyranet, line 5:
```suggestion
global.fail ??= fail;
```

tests/jest.setup.ts
```diff
@@ -0,0 +1,5 @@
+function fail(reason = 'Expected to throw, but failed to do so') {
```
kyranet, line 1:
```suggestion
function fail(reason = 'fail was called in a test.') {
```

tests/common/macros/comparators.ts
```diff
@@ -29,7 +29,7 @@ export function expectModifiedClonedValidator<T>(expected: BaseValidator<T>, act
 export function expectError<T = any>(cb: () => T, expected: BaseError) {
 	try {
 		cb();
-		fail('expected to throw, but failed to do so');
+		fail();
```
kyranet, line 32: This should be moved outside the try/catch block, and be called with `'Expected to throw, but failed to do so'`

kyranet: A `return;` at the end of the catch block allows us to not call `fail()` if the callback threw.

kyranet (changes requested):

tests/validators/string.test.ts
```diff
@@ -112,12 +112,29 @@ describe('StringValidator', () => {
...
+				'short-account-name@domain-name-that-has-more-than-sixty-three-characters-and-is-then.followed-by-another-segment-that-got-split-by-a-full-stop-symbol.com',
+				`foo@bar.${'a'.repeat(64)}`
+			])('GIVEN %p THEN throws a ConstraintError', (input) => {
 				expectError(
 					() => emailPredicate.parse(input),
 					new ExpectedConstraintError('s.string.email', 'Invalid email address', input, 'expected to be an email address')
 				);
 			});
+
+			test.each([undefined, null])('GIVEN %p THEN throws a ValidationError', (input) => {
```
kyranet, line 135: This doesn't touch the constraint, it needs to be in the validator's parse tests, at line 11.

kyranet (changes requested):
Just this one and LGTM.

kyranet (approved):

vladfrangu (approved):

PR #108, opened 2022-05-14
docs(readme): clarify the difference between validations and schemas and add table of contents

favna:


README.md
```diff
@@ -231,7 +295,144 @@ const dish = s.tuple([
...
+You can use `s.instance(Class)` to check that the input is an instance of a class. This is useful to validate inputs against classes:
+
+```typescript
+class User {
+	public constructor(public name: string) {}
+}
+
+const userInstanceValidation = s.instance(User);
+userInstanceValidation.parse(new User('Sapphire')); // => User { name: 'Sapphire' }
+userInstanceValidation.parse('oops' as any); // => throws ValidatorError
```
vladfrangu, line 354: The as any is not needed as parse accepts anything already

README.md
```diff
@@ -231,7 +295,144 @@ const dish = s.tuple([
...
+```
+
+#### Records
+
+[Back to top][toc]
+
+Record validations are similar to objects, but validate `Record<string, T>` types, keep in mind this does not check for the keys, and cannot support validation for specific ones:
```
vladfrangu, line 361:
```suggestion
Record validations are similar to objects, but validate `Record<string, T>` types. Keep in mind this does not check for the keys, and cannot support validation for specific ones:
```

vladfrangu (changes requested):

vladfrangu (approved):

PR #125, opened 2022-06-08
feat: add disabling of validators

vladfrangu:


src/index.ts
```diff
@@ -1,7 +1,7 @@
+import { getGlobalValidationEnabled, setGlobalValidationEnabled } from './lib/configs';
```
kyranet, line 1: Why is it not `export *`?

src/validators/BaseValidator.ts
```diff
@@ -61,11 +63,37 @@ export abstract class BaseValidator<T> {
...
+	 */
+	public setValidationEnabled(isValidationEnabled: boolean | null): this {
+		const clone = this.clone();
+		clone.isValidationEnabled = isValidationEnabled;
+		return clone;
+	}
+
+	public getValidationEnabled() {
+		return this.isValidationEnabled;
+	}
```
kyranet, line 91: Should be a getter, and account for global validation enabled. Then you can use this instead of two branches in every validator.

kyranet (changes requested):

src/validators/BaseValidator.ts
vladfrangu: Would be smart but this is supposed to represent the validators setting, not if the validator can run. That aside I could just make it a secondary getter

src/validators/BaseValidator.ts
```diff
@@ -90,6 +86,14 @@ export abstract class BaseValidator<T> {
 		return this.isValidationEnabled;
 	}
 
+	protected get shouldValidatorRunConstraints() {
```
kyranet, line 89: This should be named `shouldRunConstraints`, and probably type the return to `boolean` just in case?

src/validators/BaseValidator.ts
```diff
@@ -90,6 +86,14 @@ export abstract class BaseValidator<T> {
...
+		if (this.isValidationEnabled === null) {
+			return getGlobalValidationEnabled();
+		}
+
+		return this.isValidationEnabled;
```
kyranet, line 94: Could use a ternary.

kyranet (changes requested):

favna (commented):
Only Kyra's comments other than that lgtm

src/validators/BaseValidator.ts
```diff
@@ -61,11 +63,37 @@ export abstract class BaseValidator<T> {
...
+		clone.isValidationEnabled = isValidationEnabled;
+		return clone;
+	}
+
+	public getValidationEnabled() {
+		return this.isValidationEnabled;
+	}
+
+	protected get shouldRunConstraints(): boolean {
+		return this.isValidationEnabled === null ? getGlobalValidationEnabled() : this.isValidationEnabled;
```
kyranet, line 90:
```suggestion
		return this.isValidationEnabled ?? getGlobalValidationEnabled();
```

kyranet (changes requested):

kyranet (approved):

favna (approved):

PR #126, opened 2022-06-10
test: migrate to vitest

favna:
@kyranet please do custom matches ktxhbai


kyranet:
The errors were caused here: https://github.com/vitest-dev/vitest/blob/3199dc7d689a58033336bdd99075c0833d39aa81/packages/vitest/src/runtime/suite.ts#L195-L202

Which calls this: https://github.com/vitest-dev/vitest/blob/3199dc7d689a58033336bdd99075c0833d39aa81/packages/vitest/src/utils/base.ts#L89-L97

When doing `test.each([null])()`, the value, `i`, which is `null`, is converted to an empty array (`[]`) rather than being wrapped inside an array (`[null]`), which later, when spreading, caused the parameters to be... empty, which in turn makes the value `undefined`.

This is a gotcha that deviates from Jest, not sure if it was meant to align with Chai (although I don't find `each` in it), it's something Vitest's developers thought sensible, or they were simply not aware of this.

.github/workflows/continuous-integration.yml
```diff
@@ -30,6 +30,7 @@ jobs:
     name: Unit Tests
     runs-on: ubuntu-latest
     strategy:
+      fail-fast: false
```
vladfrangu, line 33: I dunno, I'd say we can keep fail-fast (if something breaks on nodejs 14 we don't need nodejs 16 or 18 to pass)

package.json
```diff
@@ -20,8 +20,8 @@
 		"lint": "eslint src tests --ext ts --fix",
 		"format": "prettier --write \"{src,tests}/**/*.ts\"",
 		"docs": "typedoc",
-		"test": "jest",
-		"test:watch": "jest --watch",
+		"test": "vitest run",
+		"test:watch": "vitest",
```
vladfrangu, line 24: Does this do watching too?

vladfrangu (approved):

package.json
favna: `vitest run` is no watch
`vitest` is watch

That said, I also recommend https://marketplace.visualstudio.com/items?itemName=ZixuanChen.vitest-explorer

.github/workflows/continuous-integration.yml
favna: The reason I prefer fail-fast false is because I want to know whether Node 16 and Node 18 actually also work or if they also fail. With fail-fast set to true you are left completely in dark and you may end up looking for a fix in completely the wrong direction.

PR #128, opened 2022-06-17
fix: compile for es2020 instead of es2021

vladfrangu:
Allows users of nodejs 14+ to use shapeshift


kyranet (changes requested):
This needs updating too:

https://github.com/sapphiredev/shapeshift/blob/a23d15f533a234a59af54a9ed60bec778f069af8/package.json#L74

favna (approved):

PR #131, opened 2022-06-19
feat: allow passing functions in `setValidationEnabled`

imranbarbhuiya:


tests/lib/configs.test.ts
```diff
@@ -11,6 +11,17 @@ describe('Validation enabled and disabled configurations', () => {
 	const setPredicate = s.set(s.number);
 	const tuplePredicate = s.tuple([s.string, s.number]);
 
+	const predicateAndValues = [
```
c43721, line 14: this looks like a minor side change, but I would suggest to do this in a separate PR to "reduce scope", but that's just me.

src/validators/BaseValidator.ts
```diff
@@ -76,18 +77,18 @@ export abstract class BaseValidator<T> {
 	 * Sets if the validator should also run constraints or just do basic checks.
 	 * @param isValidationEnabled Whether this validator should be enabled or disabled. Set to `null` to go off of the global configuration.
 	 */
-	public setValidationEnabled(isValidationEnabled: boolean | null): this {
+	public setValidationEnabled(isValidationEnabled: boolean | (() => boolean) | null): this {
 		const clone = this.clone();
 		clone.isValidationEnabled = isValidationEnabled;
 		return clone;
```
c43721, line 83: I would also add onto the doc comment that this accepts a function that returns a boolean.

tests/lib/configs.test.ts
imranbarbhuiya: I'm using this array in my new test so I don't think it's out of scope.

favna (commented):
I don't particularly mind the change but what is the use case of this exactly?

src/validators/BaseValidator.ts
favna: 🖕

c43721: 👆this is the righ emoji

imranbarbhuiya:
> I don't particularly mind the change but what is the use case of this exactly?

If you check my djs pr, I'm adding `setValidatorEnabled` every time before parsing. By allowing passing function, I can add it when defining the predicate and not every time before parsing.

vladfrangu:
> By allowing passing function, I can add it when defining the predicate and not every time before parsing.

Not really true... it'll call the function when the setValidationEnabled is called. You probably want to set the internal field to the function instead and have it called in the getter

src/validators/BaseValidator.ts
```diff
@@ -76,18 +77,18 @@ export abstract class BaseValidator<T> {
...
 	}
 
 	public getValidationEnabled() {
-		return this.isValidationEnabled;
+		return getValue(this.isValidationEnabled);
 	}
 
 	protected get shouldRunConstraints(): boolean {
-		return this.isValidationEnabled ?? getGlobalValidationEnabled();
+		return getValue(this.isValidationEnabled) ?? getGlobalValidationEnabled();
```
vladfrangu, line 91:
```suggestion
		return this.isValidationEnabled === null ? getGlobalValidationEnabled() : getValue(this.isValidationEnabled) ?? getGlobalValidationEnabled();
```

vladfrangu (changes requested):

src/validators/BaseValidator.ts
imranbarbhuiya: if `isValidationEnabled` is null, `getValue(this.isValidationEnabled)` will return null so it'll go to `getGlobalValidationEnabled()` so do we need an extra check?

vladfrangu: oh right.. derpy moments

imranbarbhuiya:
> 

Ig I'm doing the same

vladfrangu (dismissed):

vladfrangu (approved):

favna (approved):

PR #140, opened 2022-07-02
fix: return array-validator from length* methods

imranbarbhuiya:


vladfrangu (dismissed):

favna (changes requested):
Code looks good but can you add a unit test to cover `lengthGreaterThanOrEqual` in combination with `lengthLessThanOrEqual` to ensure that this doesn't break in the future?

favna (approved):

PR #141, opened 2022-07-02
feat(arrays): add unique

imranbarbhuiya:


src/validators/ArrayValidator.ts
```diff
@@ -95,6 +110,26 @@ export class ArrayValidator<T> extends BaseValidator<T[]> {
...
+	private handleUniqueStrategy(givenValues: unknown[]) {
+		if (givenValues.length < 2) return [];
+		const errors: [number, BaseError][] = [];
+		const values = [...givenValues];
+		let previousValue = values.shift();
+
+		for (let i = 0; i < values.length; i++) {
+			const value = values[i];
+			if (!Object.is(previousValue, value))
+				errors.push([i, new ValidationError('s.array(T).unique', 'Expected all values are unique', givenValues)]);
```
vladfrangu, line 123:
```suggestion
				errors.push([i, new ValidationError('s.array(T).unique', 'Expected all values to be unique', givenValues)]);
```

src/validators/ArrayValidator.ts
```diff
@@ -95,6 +110,26 @@ export class ArrayValidator<T> extends BaseValidator<T[]> {
...
+			previousValue = value;
+		}
+		return errors;
+	}
+}
+
+export const enum ArrayValidatorStrategy {
```
vladfrangu, line 130:
```suggestion
export enum ArrayValidatorStrategy {
```

tests/validators/array.test.ts
```diff
@@ -177,6 +177,20 @@ describe('ArrayValidator', () => {
 		});
 	});
 
+	describe('Unique', () => {
+		const uniquePredicate = predicate.unique;
+
+		test('GIVEN a non-unique array THEN throws CombinedPropertyError', () => {
```
vladfrangu, line 183: This isn't right... This should not throw, but an array of `['Hello', 'Hello']` should

src/validators/ArrayValidator.ts
```diff
@@ -95,6 +110,26 @@ export class ArrayValidator<T> extends BaseValidator<T[]> {
...
+
+	private handleUniqueStrategy(givenValues: unknown[]) {
+		if (givenValues.length < 2) return [];
+		const errors: [number, BaseError][] = [];
+		const values = [...givenValues];
+		let previousValue = values.shift();
+
+		for (let i = 0; i < values.length; i++) {
+			const value = values[i];
+			if (!Object.is(previousValue, value))
```
vladfrangu, line 122: Did you mean to negate this?

vladfrangu (changes requested):

src/validators/ArrayValidator.ts
```diff
@@ -95,6 +110,26 @@ export class ArrayValidator<T> extends BaseValidator<T[]> {
...
 			: Result.err(new CombinedPropertyError(errors));
 	}
+
+	private handleUniqueStrategy(givenValues: unknown[]) {
+		if (givenValues.length < 2) return [];
+		const errors: [number, BaseError][] = [];
+		const values = [...givenValues];
+		let previousValue = values.shift();
+
+		for (let i = 0; i < values.length; i++) {
```
favna, line 120: Rather than doing this you can simply do:
```ts
const givenValues: unknown[] = [];

const possiblyUnique = new Set(givenValues);

if (possiblyUnique.size === givenValues.length) {
	// is unique
} else {
	// is not unique
}
```
Because `Set`s always eliminate all non-unique values when its size matches that of the input, the input was unique.

favna (changes requested):

src/validators/ArrayValidator.ts
favna: See also comment below about using Set instead of this loop

src/validators/ArrayValidator.ts
imranbarbhuiya: If I use `Set`, I need to stringify it like [sapphiredev/shapeshift@`165b38b` (#141)](https://github.com/sapphiredev/shapeshift/pull/141/commits/165b38bc2d02f1afcae8e0990403f0f89dc03c04) this which isn't 100% accurate too. So I've reverted it to use `isDeepStrictEqual`. We can also use `fast-deep-equal` which is fater that `isDeepStrictEqual` lmk wdyt?

favna: Honestly the published version of fast-deep-equal is one of those typical very old poorly maintained packages that doesn't get releases "because it just works", for example see an issue like https://github.com/epoberezkin/fast-deep-equal/issues/43, and there are also issues such as https://github.com/epoberezkin/fast-deep-equal/issues/81.

Lastly they also ship React comparisons which is wholly unnecessary for us and would just add package bloat. 

Due to these reasons I think we should stick with `isDeepStrictEqual`.

src/validators/ArrayValidator.ts
```diff
@@ -95,6 +106,17 @@ export class ArrayValidator<T> extends BaseValidator<T[]> {
...
+	private isUnique(givenValues: unknown[]) {
+		if (givenValues.length < 2) return true;
+
+		for (let index = 1; index <= givenValues.length; index++) {
+			const previousValue = givenValues[index - 1];
+			if (isDeepStrictEqual(previousValue, givenValues[index])) return false;
+		}
+
+		return true;
+	}
```
favna, line 119: @vladfrangu  @kyranet can you guys sanity check this change I made? Tests pass as expected.

imranbarbhuiya: previous clone and shift were unnecessary. Idk why I had cloned it 😄

vladfrangu: This won't work if you do `[1, 2, 3, 2]` will it?

src/validators/ArrayValidator.ts
```diff
@@ -96,4 +109,12 @@ export class ArrayValidator<T extends unknown[], I = T[number]> extends BaseVali
...
 			: Result.err(new CombinedPropertyError(errors));
 	}
+
+	private isUnique(values: unknown[]) {
+		if (values.length < 2) return true;
+
+		const uniqueArray = uniqWith(values, isDeepStrictEqual);
+
+		return uniqueArray.length === values.length;
+	}
```
kyranet, line 119: I'm going to point this out.

Why is this not a constraint? Validators have exactly one task, and should stick to it. Extra steps must run as constraints, as any extra logic in the validators bloat the classes and make them a lot slower.

Also, that code is unworldly inefficient, to start with, it doesn't short-circuit.

kyranet (dismissed):

favna (approved):

src/validators/ArrayValidator.ts
favna: For the so manieth time, if you want better code then come up with a way for hashing to happen synchronously.

src/constraints/util/isUnique.ts
```diff
@@ -0,0 +1,9 @@
+import { isDeepStrictEqual } from 'node:util';
+import uniqWith from 'lodash.uniqwith';
+
+export const isUnique = (input: unknown[]) => {
```
vladfrangu, line 4:
```suggestion
export function isUnique(input: unknown[]) {
```

tests/validators/array.test.ts
```diff
@@ -192,6 +192,35 @@ describe('ArrayValidator', () => {
...
+			[[['Hello'], ['Hi']], s.string.array.array.unique],
+			[[[{ name: 'Hello' }], [{ name: 'Hi' }]], s.object({ name: s.string }).array.array.unique]
+		];
+
+		const invalidInputPredicate: [unknown, ArrayValidator<unknown[]>][] = [
+			[['Hello', 'Hello'], s.string.array.unique],
+			[[1, 2, 4, 2, 1], s.number.array.unique],
+			[[{ name: 'Hello' }, { name: 'Hello' }], s.object({ name: s.string }).array.unique],
+			[[['Hello'], ['Hello']], s.string.array.array.unique],
+			[[[{ name: 'Hello' }], [{ name: 'Hello' }]], s.object({ name: s.string }).array.array.unique]
```
vladfrangu, line 209: Array array?

vladfrangu (changes requested):

tests/validators/array.test.ts
vladfrangu: Oh my eyes totally skipped the non array array test too WOOPS

favna:
> Can you also just make sure this bundles correctly for browsers? (ideally it should put lodash.uniqwith in the bundle) 🙏

barring the usages of node's `inspect` the repo will now (again) be browser ready after this PR. Title has been adjusted to reflect this.

.github/workflows/continuous-integration.yml
```diff
@@ -71,6 +71,33 @@ jobs:
...
       - name: Typecheck And Build Code
         run: yarn typecheck && yarn build
+      - name: Upload bundle to artifacts
+        uses: actions/upload-artifact@3cea5372237819ed00197afe530f5a7ea3e805c8 # tag=v3
+        with:
+          name: bundle
+          path: dist/
+          if-no-files-found: error
+
+  NoModulesTesting:
```
vladfrangu, line 81: We should draft a PR for browser testing too soon

src/constraints/util/isUnique.ts
```diff
@@ -0,0 +1,8 @@
+import fastDeepEqual from 'fast-deep-equal';
```
vladfrangu, line 1: We probably wanna use the /es6 export instead, as it also handles maps, sets, etc

vladfrangu (changes requested):

src/constraints/util/isUnique.ts
favna: done

vladfrangu (approved):

PR #146, opened 2022-07-06
fix: fixed `shape#array` types

imranbarbhuiya:


vladfrangu (dismissed):

src/lib/Shapes.ts
```diff
@@ -94,7 +94,7 @@ export class Shapes {
 		return new UnionValidator(validators);
 	}
 
-	public array<T>(validator: BaseValidator<T>) {
+	public array<T extends unknown[]>(validator: BaseValidator<T[number]>) {
```
kyranet, line 97: What does this even fix? Can we please get unit tests?

kyranet (changes requested):

src/lib/Shapes.ts
imranbarbhuiya: https://github.com/discordjs/discord.js/pull/8242

iCrawl: Should probably still add a test case if you can

favna (approved):

favna:
@sapphiredev pack

vladfrangu (approved):

PR #148, opened 2022-07-06
feat: reshape finally

imranbarbhuiya:


imranbarbhuiya:
after reading readme again, i think I've understood reshape incorrectly. Will change it after dinner

tests/validators/base.test.ts
```diff
@@ -141,6 +141,29 @@ describe('BaseValidator', () => {
 		});
 	});
 
+	describe('Reshape', () => {
+		const predicate = s.string.reshape((value) => {
+			return value.length > 5
+				? Result.ok(value.length)
+				: // FIXME: if I don't add this number generic, then `reshape` isn't able to infer the type
+				  // Lmk should I keep it like this or r there any fix possible
```
favna, line 149: @kyranet @vladfrangu

imranbarbhuiya: shapeshift don't uses `@sapphiredev/result`

vladfrangu: oh lord. Maybe that's why?

imranbarbhuiya: Is it worth adding the full package when shapeshift only uses `Ok` and `err`?

favna: The idea was for shapeshift to be dependency-less, we've already had to accept some dependencies for the sake of browser compatibility / other validation library parity in `fast-deep-equal` and `lodash.uniqwith` but I would still prefer to keep it at a very low minimum.

favna (approved):
Personally I think it's fine this way. I think the way the types are set up now will be pretty seamless for the end-users and ultimately that's what matters most.

favna (approved):

vladfrangu (approved):

PR #150, opened 2022-07-07
fix: ensure browser compatibility

imranbarbhuiya:


package.json
```diff
@@ -118,5 +117,9 @@
 		"minimist": "^1.2.6"
 	},
 	"prettier": "@sapphire/prettier-config",
-	"packageManager": "yarn@3.2.1"
+	"packageManager": "yarn@3.2.1",
+	"dependencies": {
```
vladfrangu, line 121: Why are you reverting this?

.github/workflows/continuous-integration.yml
```diff
@@ -71,33 +71,6 @@ jobs:
...
       - name: Typecheck And Build Code
         run: yarn typecheck && yarn build
-      - name: Upload bundle to artifacts
-        uses: actions/upload-artifact@3cea5372237819ed00197afe530f5a7ea3e805c8 # tag=v3
-        with:
-          name: bundle
-          path: dist/
-          if-no-files-found: error
-
-  NoModulesTesting:
```
vladfrangu, line 81: Why is this reverter

vladfrangu (changes requested):

package.json
imranbarbhuiya: why do we need to bundle the deps for nodejs users? they can always install it.

.github/workflows/continuous-integration.yml
imranbarbhuiya: If we don't bundle the deps `lodash` and `fast-deep-equal` then this test won't work

package.json
vladfrangu: Makes our lives easier when bundling to have consistency

tsup.config.ts
```diff
@@ -1,18 +1,20 @@
 import { defineConfig } from 'tsup';
+import NodeModulesPolyfills from '@esbuild-plugins/node-modules-polyfill';
```
favna, line 2:
```suggestion
import { NodeModulesPolyfills } from '@esbuild-plugins/node-modules-polyfill';
```

favna (changes requested):
I would like to see

- [ ] A browser test using vitest happy-dom. This means excluding it from the normal test suite and creating a folder `browser-tests` then creating a separate vitest config and including only that folder. 
- [ ] A test for which new directory is created, a fresh node_modules is installed with only shapeshift and then an assert test is ran similar to the one deleted. Add a dependency which is a relative path to a .tgz file which is created with yarn pack in the pipeline. Then run yarn install in that directory and run the test file. Run this locally to make sure no node_modes from the root directory are loaded, I.e. by deleting it and then installing in the subdirectory.

imranbarbhuiya:
finally green ci 😄

package.json
```diff
@@ -29,39 +29,40 @@
...
 		"check-update": "cliff-jumper --dry-run",
 		"postinstall": "husky install .github/husky",
-		"prepack": "rollup-type-bundler -e node:util && pinst --disable",
+		"prepack": "rollup-type-bundler -e util && pinst --disable",
 		"postpack": "pinst --enable"
 	},
 	"devDependencies": {
 		"@commitlint/cli": "^17.0.3",
 		"@commitlint/config-conventional": "^17.0.3",
+		"@esbuild-plugins/node-modules-polyfill": "^0.1.4",
```
iCrawl, line 38: Should maybe try `brode` instead https://github.com/geut/brode/tree/main/packages/esbuild-plugin-brode
It has support for the `node:` protocol, so you wouldn't even have to touch any imports.

tsconfig.typecheck.json
```diff
@@ -2,8 +2,8 @@
 	"extends": "./tsconfig.base.json",
 	"compilerOptions": {
 		"noEmit": true,
-		"lib": ["DOM", "ESNext"],
-		"types": ["vitest/globals"]
+		"types": ["vitest/globals"],
+		"lib": ["dom", "esnext"]
```
iCrawl, line 6: Unnecessary? lol

package.json
imranbarbhuiya: as discussed in discord, it's not working in windows so windows users (contributors) won't be able to build it. Also after trying in wsl even using 
```js

			injectGlobal: {
				process: false,
				buffer: false,
				assert: false,
				crypto: false,
				console: false,
				constants: false,
				events: false,
				fs: false,
				module: false,
				path: false,
				os: false,
				url: false,
				util: true,
				timers: false,
				tty: false,
				stream: false,
				zlib: false
			}
```
it's still bundling all the core modules and making the browser build 10k+ loc. it's also adding these module imports in cjs and mjs files. I'm making this draft now and will try some other alternatives

tsconfig.typecheck.json
imranbarbhuiya: ik i'll revert it. It was bcz I was testing `happy-dom` and I added `"lib": ["dom", "esnext"]` in one `tsconfig` and after removing that deps, I tought I added `lib` here too so I removed it but then saw tests are failing bcz it was present earlier so added back lol

favna: Entire file is deleted in favour of having the few required compiler options in `tsconfig.eslint.json`. They don't affect eslint in any way and this saves on a file in the root folder.

package.json
favna: For windows compatibility best I can tell is to implement `environmentsFilter` which defaults to a noop but that doesn't change the bundle size problem. Using `@esbuild-plugins/node-modules-polyfill` the `index.global.js` is 107KB compared to the 77KB of `index.js` and `index.mjs`, a 20 KB uptick to bundle `fast-deep-equal`, `lodash.uniqwith` and `util.inspect` is pretty acceptable.

favna:
> for the 2nd point, I've removed bundling for cjs and mjs. Now it only bundles `iife`. So do u still need this test? it won't work without the other 2 deps

No more need indeed because their compatibility now gets covered again by vitest.

package.json
favna: I disagree. Dependencies should be installed by end-users for CJS / ESM. Not to mention that `fast-deep-equal` is already likely a dependency of 9/10 projects anyway as per how it is required by other packages:

image

favna (approved):

package.json
iCrawl: Yeah I can see that now, interesting. I'll see if something can be done about this some time.
Would be nice to not have to remove the `node:` protocol in source files for example.

imranbarbhuiya: I've created a pr to support `node:` protocol https://github.com/remorses/esbuild-plugins/pull/20 but idk if that repo is still maintained. So I tried https://github.com/imranbarbhuiya/esbuild-plugins-node-modules-polyfill but for some reason it is saying that `inspect` isn't exported from `util` package 😞

imranbarbhuiya: fixed in https://github.com/sapphiredev/shapeshift/pull/150/commits/47cd7c25fc11115631b9d01dda1482d175f00dc0 with a patch in the package

favna (approved):

vladfrangu (approved):

PR #152, opened 2022-07-10
build: use tsup `dts` option

favna:
Seeing as our tests do not need composite builds we can disable that in src and then enable `dts` in tsup to make the build process quicker and easier


vladfrangu (approved):
Neat. Can we apply this everywhere else too sometimes?

PR #156, opened 2022-07-10
refactor(Errors): pass error message as reason if only 1 error is received

ImRodry:


imranbarbhuiya:
You get all the errors in the error body. And if u r saying that when u log `error.message`, u want to see the first error message, then instead of `error.message`, do a `instanceOf` check and if it's `CombinedError`, then log `error.errors[0].message`

src/lib/errors/CombinedError.ts
```diff
@@ -5,7 +5,7 @@ export class CombinedError extends BaseError {
 	public readonly errors: readonly BaseError[];
 
 	public constructor(errors: readonly BaseError[]) {
-		super('Received one or more errors');
+		super(errors.length === 1 ? errors[0].message : 'Received multiple errors');
```
imranbarbhuiya, line 8:
```suggestion
		super(errors.length === 1 ? errors[0].message : `Received ${errors.length} errors`);
```
and if u r changing it then we can directly pass the number now

imranbarbhuiya (changes requested):

ImRodry:
> You get all the errors in the error body. And if u r saying that when u log `error.message`, u want to see the first error message, then instead of `error.message`, do a `instanceOf` check and if it's `CombinedError`, then log `error.errors[0].message`

The point is to get the error messages when you don't log anything (e.g. in a crash log)

imranbarbhuiya:
adding all the messages is a good improvement specially when u only use `error.message` or `error.stack` but when u log only `error`, you'll see these error messages repeating. And specially when you'll have a large schema with multiple errors.

imranbarbhuiya:
adding all the messages is a good improvement specially when u only use `error.message` or `error.stack` but when u log only `error`, you'll see these error messages repeating. And specially when you'll have a large schema with multiple errors.

imranbarbhuiya:
> For the second time, I'm making this for situations where you don't log anything because you're not expecting an error and you receive something like this

got it now. It's the issue when you get a combined error inside of another combined error and I was testing with validation error inside of combined error lol

favna:
As made quite thoroughly clear in DiscordJS "Improve errors from shapeshift" thread this PR is not to our liking and does not solve the problem of our error logging in a way that we desire. For that reason I am rejecting this PR and closing it.

ImRodry:
It solves the only issue it aims to tackle: making the error message useful. I never proposed to do anything else with this PR so I don't understand why you're saying this doesn't "solve the problem of your error logging", I believe that should be done in a separate PR.

favna:
> It solves the only issue it aims to tackle: making the error message useful. I never proposed to do anything else with this PR so I don't understand why you're saying this doesn't "solve the problem of your error logging", I believe that should be done in a separate PR.

The discussion on this topic is on Discord. If you're going to force cross-pollinating the discussion then GitHub has this beautiful "block user from org" button that we can use to force your hand.

PR #159, opened 2022-07-13
build: fix djs/builder type error

imranbarbhuiya:
similar to https://github.com/sapphiredev/shapeshift/pull/153
this pr tried to fix the djs/builder error.
djs/builder was compiled with an older version of `shapeshift` when `ArrayValidator` was accepting any type as the first generic type. But after https://github.com/sapphiredev/shapeshift/pull/140, it expects  `unknown[]` type. That's why users are getting type errors. 

> **Note**
`type makeArray<T> = T extends any[] ? T : T[];` can be removed by making `P` 2nd generic type instead of 3rd. But then it'll throw the same error for packages that were compiled with a new version of `shapeshift` which's 2nd param is not an array type.
So in the future, when we'll release v4, we can remove it

### DIFF
--- a/src/validators/ArrayValidator.ts
+++ b/src/validators/ArrayValidator.ts
@@ -20,6 +20,7 @@ import { BaseValidator } from './imports';
 
 type makeArray<T> = T extends any[] ? T : T[];
 
+// TODO(v4): revert https://github.com/sapphiredev/shapeshift/pull/159
 export class ArrayValidator<T, I = makeArray<T>[number], P extends unknown[] = makeArray<T>> extends BaseValidator<P> {
 	private readonly validator: BaseValidator<I>;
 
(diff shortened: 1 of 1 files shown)


src/validators/ArrayValidator.ts
```diff
@@ -18,78 +18,80 @@ import { Result } from '../lib/Result';
 import type { ExpandSmallerTuples, Tuple, UnshiftTuple } from '../lib/util-types';
 import { BaseValidator } from './imports';
 
-export class ArrayValidator<T extends unknown[], I = T[number]> extends BaseValidator<T> {
+type makeArray<T> = T extends any[] ? T : T[];
+
+export class ArrayValidator<T, I = makeArray<T>[number], P extends unknown[] = makeArray<T>> extends BaseValidator<P> {
```
favna, line 23: Please add the TODO from the PR description here in the code as well with: `// TODO (v4): `

favna (changes requested):

favna (approved):

imranbarbhuiya:
I just noticed that if anyone has `ArrayValidator<string[]>` then
without this pr,  return type of `.parse` will be `string[]`
with this pr, it'll be `string`. 

So we are again breaking. So I'm closing this pr and let's wait for a new builder release. 
Anyone who has the issue they either need to add a resolution of `^3.4.0` or `skipLibCheck` before /builder releases a new version.

Issue #200, opened 2022-09-29
request: add `s.string.phone` to regex validate a phone number


favna:
### Is there an existing issue or pull request for this?

- [X] I have searched the existing issues and pull requests

### Feature description

We have `s.string.email` and in forms it is often required to enter both an email and a phone number so lets also add `s.string.phone`

### Desired solution

- Add a new validation `s.string.phone`
- Add unit tests

### Alternatives considered

Using `s.string.regex` with the regex for a phone number, however this is not ideal because then end-users have to constantly look up the phone regex online

### Additional context

Refer to https://ihateregex.io/expr/phone/ for an example for a phone regex

imranbarbhuiya:
I'm working on it

PR #202, opened 2022-09-30
feat: add `s.string.phone` 

imranbarbhuiya:
fixes: #200

### DIFF
--- a/tests/validators/string.test.ts
+++ b/tests/validators/string.test.ts
@@ -392,7 +392,7 @@ describe('StringValidator', () => {
 		});
 
 		describe('phone', () => {
-			const phonePredicate = s.string.phone();
+			const phonePredicate = s.string.phone;
 
 			test.each(['+79919542975', '786-307-3615', '+16308520397', '+919367788755', '+916000060091', '9365706789', '936-570-6789'])(
 				'GIVEN %j THEN returns given value',
--- a/tests/validators/string.test.ts
+++ b/tests/validators/string.test.ts
@@ -365,31 +365,51 @@ describe('StringValidator', () => {
 					);
 				});
 			});
+		});
+		describe('date', () => {
+			const stringDatePredicate = s.string.date;
 
-			describe('date', () => {
-				const stringDatePredicate = s.string.date;
-
-				test.each([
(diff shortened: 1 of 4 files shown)


favna (approved):

tests/validators/string.test.ts
```diff
@@ -365,31 +365,51 @@ describe('StringValidator', () => {
...
 
-				test('GIVEN invalid date string THEN throws a ConstraintError', () => {
+			test.each(['+79919542975', '786-307-3615', '+16308520397', '+919367788755', '+916000060091', '9365706789', '936-570-6789'])(
+				'GIVEN %j THEN returns given value',
+				(input) => {
+					expect(phonePredicate.parse(input)).toBe(input);
+				}
+			);
+
+			test.each(['+1 555-555-5555 ext', '+1 (555) 555-5555', '987-123-4567 x12345', '(123) 456-7890 ext12345'])(
```
vladfrangu, line 404: Why does `+1 (555) 555-5555` fail exactly? it *is* a valid number 🤔

vladfrangu (changes requested):

tests/validators/string.test.ts
imranbarbhuiya: I used the regex fava provided. I don't have much idea of numbers with area code. Perhaps u can't use area code with country code? If that's not the case, I'll search for a better regex

favna: looks like this one may be better:
```ts
^(\+\d{1,2}\s)?\(?\d{3}\)?[\s.-]?\d{3}[\s.-]?\d{4}$
```

Source: https://stackoverflow.com/a/16699507

imranbarbhuiya:
I've updated the regex. Also changed `phone` from `getter` to a `method`. The reason is maybe we can add `allowedCountryCode`. Will try to do that in a followup pr

vladfrangu:
Getting that quota are we? :P Yeah, sure a different PR for adding that would be fine (since it'd also need more testing)

vladfrangu (approved):

favna (approved):
sounds quite complex to do country code so you rather than me lol. But hey if you can figure it out then awesome.

PR #223, opened 2022-11-15
feat: add `when` constraint

imranbarbhuiya:
~~_Not ready for review yet. I haven't even started. Just fixed the typings and added a test_~~

~~I'll finish it after my exams. In the meantime, if anyone else is interested then u can create a pr and I'll close this one~~

The code is ready for review.

TODO: 
- [x] Add more tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,45 +16,59 @@ Blazing fast input validation and transformation ⚡
 
 ## Table of Contents
 
--   [@sapphire/shapeshift](#sapphireshapeshift)
-    -   [Table of Contents](#table-of-contents)
-    -   [Description](#description)
-    -   [Features](#features)
-    -   [Usage](#usage)
-        -   [Basic usage](#basic-usage)
-        -   [Defining validations](#defining-validations)
-            -   [Primitives](#primitives)
-            -   [Literals](#literals)
-            -   [Strings](#strings)
-            -   [Numbers](#numbers)
-            -   [BigInts](#bigints)
-            -   [Booleans](#booleans)
-            -   [Arrays](#arrays)
-            -   [Tuples](#tuples)
-            -   [Unions](#unions)
-            -   [Enums](#enums)
-            -   [Maps](#maps)
-            -   [Sets](#sets)
-            -   [Instances](#instances)
-            -   [Records](#records)
-            -   [Functions // TODO](#functions--todo)
-            -   [TypedArray](#typedarray)
-        -   [Defining schemas (objects)](#defining-schemas-objects)
-            -   [Utility types for TypeScript](#utility-types-for-typescript)
-                -   [Extracting an interface from a schema](#extracting-an-interface-from-a-schema)
-                -   [Defining the structure of a schema through an interface](#defining-the-structure-of-a-schema-through-an-interface)
-            -   [`.extend`:](#extend)
-            -   [`.pick` / `.omit`:](#pick--omit)
-            -   [`.partial`](#partial)
-            -   [`.required`](#required)
-        -   [Handling unrecognized keys](#handling-unrecognized-keys)
-            -   [`.strict`](#strict)
-            -   [`.ignore`](#ignore)
-            -   [`.passthrough`](#passthrough)
-        -   [BaseValidator: methods and properties](#basevalidator-methods-and-properties)
-        -   [Enabling and disabling validation](#enabling-and-disabling-validation)
-    -   [Buy us some doughnuts](#buy-us-some-doughnuts)
-    -   [Contributors](#contributors)
+- [@sapphire/shapeshift](#sapphireshapeshift)
+  - [Table of Contents](#table-of-contents)
+  - [Description](#description)
+  - [Features](#features)
+  - [Usage](#usage)
+    - [Basic usage](#basic-usage)
+    - [Defining validations](#defining-validations)
+      - [Primitives](#primitives)
+      - [Literals](#literals)
+      - [Strings](#strings)
+      - [Numbers](#numbers)
+      - [BigInts](#bigints)
+      - [Booleans](#booleans)
+      - [Arrays](#arrays)
+      - [Tuples](#tuples)
+      - [Unions](#unions)
+      - [Enums](#enums)
+      - [Maps](#maps)
+      - [Sets](#sets)
+      - [Instances](#instances)
+      - [Records](#records)
+      - [Functions // TODO](#functions--todo)
+      - [TypedArray](#typedarray)
+    - [Defining schemas (objects)](#defining-schemas-objects)
+      - [Utility types for TypeScript](#utility-types-for-typescript)
+        - [Extracting an interface from a schema](#extracting-an-interface-from-a-schema)
+        - [Defining the structure of a schema through an interface](#defining-the-structure-of-a-schema-through-an-interface)
+      - [`.extend`:](#extend)
+      - [`.pick` / `.omit`:](#pick--omit)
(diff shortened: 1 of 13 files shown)


README.md
```diff
@@ -809,12 +809,10 @@ provides the schema when `is` resolves truthy, and `otherwise` provides the sche
...
 When `is` is not provided (`=== undefined`) it is strictly resolved as `Boolean(value)` wherein `value` is the current
-value of the referenced sibling. Note that if the refeferenced sibling is an array that this will mean the result is
-always `true` because `Boolean([...])` is true.
+value of the referenced sibling. Note that if multiple siblings are referenced then all the values of the array need to
+resolve truthy for the `is` to resolve truthy.
 
-When `is` is a primitive literal it is strictly compared (`===`) to the current value of the referenced sibling, or
-through [`Array#some`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some) with
-a strict negative comparison (`!==`) for each element if multiple siblings are referenced.
+When `is` is a primitive literal it is strictly compared (`===`) to the current value.
```
favna, line 815: Why did you delete the blurb about using `Array#some`?

src/constraints/ObjectConstrains.ts
```diff
@@ -38,18 +40,14 @@ export function whenConstraint<T extends BaseValidator<any>, I, Key extends When
...
+		return isKeyArray ? !value.some((val: any) => !val) : Boolean(value);
 	}
 
 	if (typeof options.is === 'function') {
 		return options.is(value);
 	}
 
-	if (Array.isArray(value)) {
-		return !value.some((val) => val !== options.is);
-	}
```
favna: Why is this gone? This should account for `options.is` being defined `'foobar'` and checking if every value in keys `['stringLike1', 'stringLike2']` is equal to `'foobar'`

That doesn't get covered by the `.some` above which it only applies when `options.is` is `undefined`

favna (changes requested):

src/constraints/ObjectConstrains.ts
imranbarbhuiya: The check was for `Array.isArray(value)` so what'll happen is
```ts
s.object(
{ a: s.array(s.string), c: s.string.when('a', {
is: 's',
...
})
```
Here, it'll validate that the array contains only `'s'`. And I don't think we should support the array check. If we are supporting any type of object, we'll need to support all the object types (obj, map, set, ...) and that'll be out of scope for shapeshift especially when we are already providing a way (using cb) to compare using your own eq check. Also for the below case where it'll check if the array contains elements `['s']` 
```ts
s.object(
{ a: s.array(s.string), c: s.string.when('a', {
is: [ 's' ],
...
})
``` 
So then we'll need to check for deep equality of arrays too. So better just keep `===` check when `is` is provided (yup does the same ig) and if `is` isn't provided, then check 
1. if only one key is present, then `Boolean(value)` 
2. if multiple is present, `array.every(Boolean)/the one we are using`

_In my first message, I said about when `is` is undefined and multiple keys are present then check truty of each elements of the value but there was a confusion with `values` array_

README.md
imranbarbhuiya: 2nd point doesn't need it now. 1st point does use it so I just added what happens when the keys are an array. I can add this `some` part to the first point too if u want

package.json
```diff
@@ -118,10 +118,9 @@
 		"ansi-regex": "^5.0.1",
 		"minimist": "^1.2.7"
 	},
-	"prettier": "@sapphire/prettier-config",
 	"packageManager": "yarn@3.3.0",
 	"dependencies": {
 		"fast-deep-equal": "^3.1.3",
-		"lodash.uniqwith": "^4.5.0"
+		"lodash": "^4.17.21"
```
vladfrangu, line 124: Let's not include the entirety of lodash please, just the parts we need (even if they're removed from the bundle, lodash is massive)

src/constraints/ObjectConstrains.ts
```diff
@@ -0,0 +1,53 @@
...
+
+			const value = isKeyArray ? key.map((k) => get(parent, k)) : get(parent, key);
+
+			const predicate = resolveBooleanIs<T, Key>(options, value, isKeyArray) ? options.then : options.otherwise;
+
+			if (predicate) {
+				return predicate(validator).run(input) as Result<I, ExpectedConstraintError<I>>;
+			}
+
+			return Result.ok(input);
```
vladfrangu, line 38: Should it return the input if no predicate is found?

src/constraints/ObjectConstrains.ts
```diff
@@ -0,0 +1,53 @@
...
+import type { IConstraint } from './type-exports';
+
+export type ObjectConstraintName = `s.object(T.when)`;
+
+export type WhenKey = PropertyKey | PropertyKey[];
+
+export interface WhenOptions<T extends BaseValidator<any>, Key extends WhenKey> {
+	is?: boolean | ((value: Key extends Array<any> ? any[] : any) => boolean);
+	then: (predicate: T) => T;
+	otherwise?: (predicate: T) => T;
```
vladfrangu, line 14: Isn't a `when()` with just `then` do the same as a `default()`? If yes, could we make a type that somehow says "either is or otherwise must be present when using when"?

vladfrangu (changes requested):

src/constraints/ObjectConstrains.ts
favna: I don't think so? Maybe a practical example with some work code that uses Yup helps:
```ts
  depatureDate: date()
    .when([openAccordionsDepartureArrivalDateTimeKey, 'departureOrArrivalToggle'], {
      is: (openAccordionsDepartureArrivalDateTimeValue: boolean, departureOrArrivalToggleValue: DepartureOrArrivalToggleUnion) =>
        openAccordionsDepartureArrivalDateTimeValue && departureOrArrivalToggleValue === 'departure',
      then: (schema) =>
        schema
          .min(subDays(startOfDay(new Date()), 1), 'Departure date may not be more than 1 day ago.')
          .max(startOfDay(new Date('2100-01-01T00:00:00Z')), 'Departure date may not be after the year 2099.')
          .defined('Departure date is required.')
          .transform(invalidDateToNull)
          .test(
            'departure-date-is-required-when-departure-time-is-provided',
            'Departure date is required when departure time has been provided.',
            (value, context) => testDepartureArrivalDateTimeInvariableRequired(value, context, 'departureTime')
          )
    })
    .nullable(),
```

Or for another form:
```ts
  arrivalDate: date()
    .when(['departureOrArrival', 'trainIdentifications'], {
      is: (departureOrArrivalValue: Capitalize<DepartureOrArrivalToggleUnion>, trainIdentifications: SchemaRequestNewTrainIdentifications) =>
        departureOrArrivalValue === 'Arrival' && !isMultipleDaysRequest(trainIdentifications),
      then: (schema) =>
        schema
          .test('is-not-before-or-after-train-date-minus-1-day', 'Arrival date should be one day before or after the train date.', testVertrekAankomstDatumTijd)
          .transform(invalidDateToNull)
          .required('Arrival date is required.')
    })
    .nullable()
    .defined(),
```

src/constraints/ObjectConstrains.ts
imranbarbhuiya: if `otherwise` isn't present then it returns the input.

vladfrangu (dismissed):

favna (approved):

PR #253, opened 2023-03-17
fix(*): TypeScript 5.x compatibility

favna:
In particular, add compatibility with `verbatimModuleSyntax`.

Other than that I removed the last remaining `const enum` export.


favna:
@sapphiredev pack

PR #275, opened 2023-06-04
fix(arrayvalidator): fixed runaway type instantiation with TypeScript >=5.1

favna:
Related: https://github.com/microsoft/TypeScript/issues/54491

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,3 +1,11 @@
 {
-	"extends": "@sapphire"
+	"extends": "@sapphire",
+	"overrides": [
+		{
+			"files": ["tests/**/*.ts"],
+			"rules": {
+				"@typescript-eslint/dot-notation": "off"
+			}
+		}
+	]
 }
--- a/package.json
+++ b/package.json
@@ -35,8 +35,8 @@
 		"lodash": "^4.17.21"
 	},
 	"devDependencies": {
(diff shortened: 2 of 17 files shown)


favna:
@sapphiredev pack

vladfrangu (approved):
If it works, merge it

vladfrangu (approved):

SuperchupuDev:
this fix can probably be reverted now that typescript 5.1.5 was released
- https://github.com/microsoft/TypeScript/issues/54491

favna:
> this fix can probably be reverted now that typescript 5.1.5 was released

Fair point but because it's still an issue in older versions of TypeScript and we cannot be guaranteed that everyone uses caret versioning, or that their transitive dependencies are properly updated, we will keep the fix for now.

SuperchupuDev:
alright but 5.1.3 is the only stable typescript version that had the problem, there are no other versions with the issue, updating to a new patch release shouldn't be that much of an problem? specially to fix a regression

imranbarbhuiya:
Why does this pr need to revert tho? are there any issues with the current approach?

vladfrangu:
Unless there are any glaring issues caused by this PR, we won't be reverting it any time soon, as (as far as we know) there are no issues with keeping it. Should that change, you're free to open an issue / revert

Issue #316, opened 2023-11-18
request: cloudflare worker support


mandar1jn:
### Is there an existing issue or pull request for this?

- [X] I have searched the existing issues and pull requests

### Feature description

From #248 I could gather that cloudflare worker support is something that was being kept in mind. I recently tried to use @discordjs/builders which depends on sapphire and it gave me errors for needing to enable node_compat, something I could gather is heavily discouraged

### Desired solution

Implement an alternative for the inspect function imports from util 

### Alternatives considered

Find another package which provides the same functionality as the inspect function

favna:
I'm not sure where you draw the conclusion from that node_compat is discouraged? The docs don't mention any such thing: https://developers.cloudflare.com/workers/runtime-apis/nodejs

SuperchupuDev:
@favna the wrangler CLI mentions that node_compat has serious tradeoffs
![image](https://github.com/sapphiredev/shapeshift/assets/53496941/e75049be-fac5-4374-9401-433c06357782)

favna:
Regarding Wrangler using code from ionic, you'll have to take that up with Cloudflare. As for the usage of `node:util` in this lib, we are very much aware of the issue, but at the same time personally, I have no idea how we can replace it. Possibly @kyranet who designed that part of the code can toss in some ideas. Alternatively, I'm not entirely sure why using @imranbarbhuiya's `esbuild-plugins-node-modules-polyfill` does not replace `var util = require('util');` / `import { inspect } from 'util';` from the CJS and ESM builds respectively when the plugin is activated for those bundles. That might be a way to achieve wrangler support as well.

That all said, I'm not sure if just replacing util is enough, because we also depend on:
```js
import get from 'lodash/get.js';
import fastDeepEqual from 'fast-deep-equal/es6/index.js';
import uniqWith from 'lodash/uniqWith.js';
```
And outside of bundling all the code into the shapeshift bundle I don't see any way to replace those.

imranbarbhuiya:
wrangler is also planning to use this plugin, maybe it'll solve the issue. see https://github.com/cloudflare/workers-sdk/pull/3832 but they want to ship it as a breaking change so I am not sure if it'll happen anytime soon

PR #334, opened 2023-12-17
chore: bundle node util in node env too

imranbarbhuiya:
There are 2 approaches, see first and second commit. In the first approach, we'll need to add the node module name every time we use a new module, in 2nd approach, we'll need to add the deps name every time we add a new deps and don't want to bundle it. 

check [discord channel](https://canary.discord.com/channels/737141877803057244/934614121687687179/1184234577439494334) for more context

fixes: #316


tsup.config.ts
```diff
@@ -6,12 +6,13 @@ const baseOptions: Options = {
 	dts: true,
 	entry: ['src/index.ts'],
 	minify: false,
-	skipNodeModulesBundle: true,
+	external: ['fast-deep-equal', 'lodash'],
```
favna, line 9:
```suggestion
	external: Object.keys(dependencies),
```

favna (changes requested):
Add to the top of the tsup config:
```
import { dependencies } from './package.json';
```

Then I think this solution is what we'll go for.

imranbarbhuiya:
I was in vacation and after that I forgot about this pr, @favna applied your suggestion in the last commit

favna (approved):

PR #341, opened 2024-01-18
fix: fixed commonjs typings export mapping

favna:
Similar to sapphiredev/plugins#537


package.json
```diff
@@ -27,7 +27,8 @@
 		"format": "prettier --write \"{src,tests}/**/*.ts\"",
 		"docs": "typedoc-json-parser",
 		"test": "vitest run",
-		"build": "tsup",
+		"build": "tsup && yarn build:rename-cjs-index",
+		"build:rename-cjs-index": "move-file dist/cjs/index.d.ts dist/cjs/index.d.cts",
```
vladfrangu, line 31: Why are we using a script in places and a module in others

vladfrangu (changes requested):

vladfrangu (approved):

PR #446, opened 2025-09-18
chore: remove rimraf, fix @types/node version

beeequeue:
This PR removes `rimraf` from the devDependencies.

I decided to just remove the `clean` script, as it's already done by the `tsup` config and it's still possible to do manually if it's ever needed.

I also moved the `resolutions` and `packageManager` fields in the manifest to the rest of the dependency related keys, so they aren't as hidden. When doing this I found that the `@types/node` version is both too high for the supported node version and overridden by the `resolutions`.
So I downgraded the non-overridden version to the latest `20.x` version and set the override to the same.


favna (approved):
lgtm but there is a merge conflict due to renovate

favna (approved):

